PR #58, opened 2012-01-14
Sort predicate_select keys based on order of :only option

ndbroadbent:
Sort predicate_select keys based on order of :only option, so that user has the ability to prioritze order of predicates.

I wanted to change the order of predicates in the select list, and was surprised that the following did not have that effect:

``` haml
f.predicate_select :only => [:cont, :not_cont, :matches, :does_not_match, :blank, :present, :eq, :not_eq], :compounds => false
```

### DIFF
--- a/lib/ransack/helpers/form_builder.rb
+++ b/lib/ransack/helpers/form_builder.rb
@@ -115,16 +115,10 @@ module Ransack
 
       def predicate_select(options = {}, html_options = {})
         options[:compounds] = true if options[:compounds].nil?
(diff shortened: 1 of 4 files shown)


avit:
This PR is getting a bit old... is it still worth refreshing it? I could rebase and update this if needed.

seanfcarroll:
Does anyone still want this? I can rebase it if we want to keep it, otherwise I think we should close it.

seanfcarroll:
Closed due to lack of interest.

Issue #271, opened 2013-08-21
Add configuration option: distinct


vemv:
I'd like the `distinct` option of the `result` to be true by default in my app. AFAICT this is not currently possible.

Could you please add that option to those accepted by Ransack.configure?

PR #274, opened 2013-09-03
Add ransack context opts config distinct true by default

momer:
Resolves issue [#271](https://github.com/ernie/ransack/issues/271) by enabling distinct: true by default for Context.evaluate


avit:
This PR is getting old: it also includes additional changes that are outside of what is described here. Do we still want this? If it's relevant I can rebase and update it.

seanfcarroll:
A SQL DISTINCT clause uses a different execution path than a regular query, often a more inefficient one. I think automatically adding a distinct to every query will potentially add overhead and reduce the performance of Ransack.

I think it would be best to retain the current behaviour, and allow the optional addition of DISTINCT when needed. 

I vote to close this without merge. Any opinions? @momer @avit @radar

seanfcarroll:
I'm voting to close this. Any opinions?

seanfcarroll:
Closing due to lack of interest.

Issue #156, opened 2012-10-31
how to use a scope for the search ?


luisfernandos:
I am updating a rails application that uses searchlogic to ransack and I need some help to create predicates. I have this 2 scopes for getting a date interval for bdays, here they are:

  scope :date_of_birth_gte, lambda {|d|
      d1 = Date.strptime(d,I18n.t("date.formats.default"))
      {
              :conditions=>["RIGHT(date_of_birth,5) >= '#{"%02d" % d1.month}-#{"%02d" % d1.day}' and date_of_birth IS NOT NULL"]
      }
    }

```
scope :date_of_birth_lte, lambda {|d|
  d1 = Date.strptime(d,I18n.t("date.formats.default"))
  {
          :conditions=>["RIGHT(date_of_birth,5) <= '#{"%02d" % d1.month}-#{"%02d" % d1.day}'  and date_of_birth IS NOT NULL"]
  }
}
```

   How can I enable ransack use such scopes. It would help a lot if somebody point some documentation; seems I need to do something in the configuration and in the model, however I have no idea from where I can start.
  Any help would be appreciated.
  Thanks a lot

Issue #61, opened 2012-01-26
Support for existing named scopes?


webervin:
Is it planned or is there already any way to use defined named scopes as  ransacker?

PR #288, opened 2013-10-15
Expose model scopes for search criteria

avit:
Using `ransackable_scopes`, users can define whitelists for allowed
model scopes.

This implementation is my +1 for issues #61 and #156.

### DIFF
--- a/lib/ransack/search.rb
+++ b/lib/ransack/search.rb
@@ -26,16 +26,12 @@ module Ransack
     end
 
     def build(params)
-      attribute_method = proc { |k| @base.attribute_method?(k) }
-      scope_method     = proc { |k| @context.ransackable_scope?(k, @context.object) }
-
       collapse_multiparameter_attributes!(params).each do |key, value|
-        case key
-        when 's', 'sorts'
+        if key == 's' || key == 'sorts'
           send("#{key}=", value)
-        when attribute_method
+        elsif @base.attribute_method?(key)
           base.send("#{key}=", value)
-        when scope_method
+        elsif @context.ransackable_scope?(key, @context.object)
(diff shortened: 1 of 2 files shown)


lib/ransack/search.rb
```diff
@@ -27,11 +27,12 @@ def result(opts = {})
...
       collapse_multiparameter_attributes!(params).each do |key, value|
-        case key
-        when 's', 'sorts'
+        if key == 's' || key == 'sorts'
           send("#{key}=", value)
-        else
-          base.send("#{key}=", value) if base.attribute_method?(key)
+        elsif @base.attribute_method?(key)
+          base.send("#{key}=", value)
+        elsif @context.ransackable_scope?(key, @context.object)
```
sineed: @avit it was a long time ago but maybe you remember why this is implemented in `Search` class? I think it will be more convenient to have this logic in `Grouping` class because it will be possible to build scopes inside groupings, e.g.  params below would be correct:

```
g: [
  { over_age: 20 },
  { name_eq: 'Denis' },
  c: [ ...]
]
```

Regarding to using OR with scopes  it still will be incorrect, but it should not break current behavior of scopes.

Please, correct me if it can't be done on `Grouping` layer for some reason

avit: That might be workable, I think I suggested [something like it][0] but it was just easier to build into the existing structure, see my [original comment][1]. Part of it was so things like ActiveAdmin and RansackUI would "just work" but maybe not the cleanest approach.

[0]: https://github.com/activerecord-hackery/ransack/pull/288#issuecomment-26762566
[1]: https://github.com/activerecord-hackery/ransack/pull/288#issuecomment-28419062

Issue #299, opened 2013-11-07
Ransack doesn't handle attribute names including "and" / "or"


seanlinsley:
Starting from: https://github.com/gregbell/active_admin/issues/2669

If you have a `foo_and_bar` attribute on a model, this should work:
```ruby
Model.search(foo_and_bar_eq: 'baz').result
```

My best guess is that right now the logic that allows this:
```ruby
User.search(first_name_or_last_name_eq: 'baz').result
```
```sql
SELECT "users".* FROM "users" WHERE (("users"."first_name" = 'baz' OR "users"."last_name" = 'baz'))
```

Is skipping past existing attributes, and likely also skipping past existing ransackers.

PR #302, opened 2013-11-13
[299] Let attribute names include "and" / "or"

seanlinsley:
This is my attempt at fixing #299. It works, but likely doesn't cover all scenarios.

The `/_and_|_or_/` style logic exists all over the place; for example I know the translation code is still broken in the context of #299. Maybe it'd be better to extract the attribute splitting logic out to a single method that all code would use.


timoschilling:
This PR is should be replaced by #562 (which is merged) and can be closed

jonatack:
Thanks @timoschilling .

Issue #46, opened 2011-12-13
simple_form instead of search_form?


ghost:
I'm using my search forms like this:

= simple_form_for @q, :url => categories_path, :html => {:method => :get} do |f|

and it works

instead of 

= search_form_for @q do |f|

There are problems if I use simple_form?

Issue #31, opened 2011-10-17
Using Ransack with Formtastic


danbee:
Hi,

How can I use formtastic to render the search form? It doesn't work if I just swap out search_form_for with semantic_form_for.

Thanks,
- Dan

PR #341, opened 2014-02-21
Combination with SimpleForm's form builder is possible now.

nagyt234:
By changing only one line of code (the class inheritance) in lib/ransack/helpers/form_builder.rb and set an environment variable is it possible to combine form builders of ransack and SimpleForm.

This could be an answer for the issues #31 and #46.


dgm:
I know this feature is old, but .. is there a reason this uses an ENV variable instead of a more standard config item?

Issue #120, opened 2012-08-06
Mongoid support?


ghost:
It would be great if this could support Mongoid.

PR #407, opened 2014-08-04
Mongoid 4.0 support

Zhomart:
issue #120

Hi, I've added support for mongoid, but without associations. 

Here what I've done:
- active_record, arel codes moved into `adapters/active_record`
  - ransack classes that hardly depend on AR/Arel are moved inside `adapters/active_record/ransack`
- mongoid codes are inside `adapters/mongoid`
- created `Attribute` class for mongoid that imitates `Arel::Attributes::Attribute`
- all existing ransack specs remained, added specs for mongoid
- disabled joins for mongoid. there are no specs for mongoid/joins
- added rake tasks
  - `$ rake` runs active_record specs. **default**
  - `$ DB=mongodb rake` runs mongoid specs

### DIFF
--- a/lib/ransack/adapters/active_record/ransack/context.rb
+++ b/lib/ransack/adapters/active_record/ransack/context.rb
@@ -28,32 +28,6 @@ module Ransack
       @join_type = options[:join_type] || Polyamorous::OuterJoin
       @search_key = options[:search_key] || Ransack.options[:search_key]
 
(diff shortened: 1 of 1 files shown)


pencilcheck:
Mongoid stores its relational and embedded attributes in class arrays, I wonder if we can take advantage of that somehow ... And if the only blocker from getting it done is join spec, but not any technical difficulties, someone knowledgeable should come up with a more detail join spec. Mongoid does some denormalization internally I believe.

Issue #472, opened 2014-11-24
id column not receiving array for custom ransacker


johnrees:
I have a ransacker called `:student_users` and use the filter `student_users_in`

I've simplified the code below to try and show what is happening

```
ransacker :student_users, :formatter => proc { |v|
  [10,11]
} do |parent|
  parent.table[:id]
end
...
SELECT DISTINCT "users".* FROM "users" WHERE "users"."id" IN (NULL)
```
I'd expect the above to return `SELECT DISTINCT "users".* FROM "users" WHERE "users"."id" IN (10,11)`

If I return an integer in the proc then it behaves as I'd expect

```
ransacker :student_users, :formatter => proc { |v|
  10
} do |parent|
  parent.table[:id]
end
...
SELECT DISTINCT "users".* FROM "users" WHERE "users"."id" IN (10)
```

If I filter the `first_name` (string) field instead of `id`, it shows that the array is actually getting passed from the proc

```
ransacker :student_users, :formatter => proc { |v|
  [10,11]
} do |parent|
  parent.table[:first_name]
end
...
SELECT DISTINCT "users".* FROM "users" WHERE "users"."first_name" IN ('---
- 10
- 11
')
```

Have I misunderstood something here?

ideanl:
I am experiencing the same issue, and have looked through the source a bit, but I'll keep digging for the bug. It appears that the first name is being passed in incorrectly, so that's also an issue right?

jonatack:
@johnrees, @ideanl, thanks for the issue.

The first step to take would be to write a test for the problem that works with Arel 5 and that fails with Arel 6 or Arel master.

This would allow others to reproduce the problem easily and then work efficiently to solve it.

jonatack:
@johnrees, @ideanl, thanks for the issue.

The first step would be to contribute a test for the problem that works with Arel 5 and that fails with Arel 6 or Arel master.

This would allow others to reproduce the problem easily and then work efficiently to solve it.

I often spend more time trying to reproduce the problem and the OP's environment than in finding the bug. Even if you don't have an answer to the problem, contributing regression specs encourages others who aren't encountering the issue to have a look with their limited time ;).

jonatack:
@johnrees, @ideanl, thanks for the issue.

The first step would be to contribute a test for the problem that works with Arel 5 and that fails with Arel 6 or Arel master.

This allows others to reproduce the issue easily and then work efficiently to solve it.

[...]

PR #486, opened 2015-01-03
[472] Allow arrays to be received for custom ransacker

ideanl:
This is a try at fixing #472 and it works for me. Arel's update now has an Arel::Nodes::Casted type instead of the original array, so NULL gets passed into the SQL statement. This should cover all cases, and an if statement has been placed so this only will apply for when the Arel::Nodes::Casted type is used.

### DIFF
--- a/lib/ransack/adapters/active_record/ransack/nodes/condition.rb
+++ b/lib/ransack/adapters/active_record/ransack/nodes/condition.rb
@@ -18,6 +18,7 @@ module Ransack
             predicates.inject(&:or)
           end
         else
+          predicates.first.right[0] = predicates.first.right[0].val if predicates.first.right[0].class == Arel::Nodes::Casted
           predicates.first
         end
       end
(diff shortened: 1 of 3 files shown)


jonatack:
Thanks, @ideanl. The test suite is not passing. Did running the Ransack specs work for you with your PR?

It seems to me that the first step to take would be to write a test for the problem that works with Arel 5 and fails with Arel 6 or Arel master. This would allow others to replicate the problem and then work efficiently to solve it.

ideanl:
Ok, I've updated the PR with a test that passes with Arel 5, but fails in Arel 6, and I've made all the tests pass by revising the code I added.

jonatack:
Thanks @ideanl, this improves compatibility with Rails 4.2 (Arel 6.0).

jonatack:
Would you like to update the change log?

It still does not work the way the OP wanted with searching on a string in #472, but yes, as you pointed out that's a separate issue.

ideanl:
PR #487 is an update to the change log, now including this fix.

jonatack:
Thanks :smiley:

PR #488, opened 2015-01-05
Allow string arrays to be received for custom ransacker.

ideanl:
This is a fix for the last example in #472 in which filtering on a string field does not work in Arel >= 6. The test written passes with Arel 5 without the fix, and the fix allows the test to pass in Arel 6.


jonatack:
Great! Thank you :tada:

PR #498, opened 2015-01-21
Embedded Relations in Mongoid

pencilcheck:
Support referenced relations for mongoid in ransack is very difficult due to the fact that mongodb has no joins, therefore I guess might as well start with something that I have a workable solution to.

I have not written any specs for this PR yet, just to throw it out there to see if this is a good idea.

### DIFF
--- a/lib/ransack/adapters/mongoid/base.rb
+++ b/lib/ransack/adapters/mongoid/base.rb
@@ -73,7 +73,7 @@ module Ransack
           end
 
           def reflect_on_all_associations_all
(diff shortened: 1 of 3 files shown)


jonatack:
Thank you @pencilcheck.

I don't have Mongoid installed right now.

If you add spec coverage, I'm :+1: with trusting your work and merging it in, if it will help people using Ransack with Mongoid.

jonatack:
Mongoid users: specs would be welcome! cc @pencilcheck

pencilcheck:
Sure, I will submit specs when I have more time. Thanks for the reminder.

PR #503, opened 2015-01-28
Don't replace 0 and 1 with false and true, fixes #502

### LINKED ISSUE #502: Wrong result and errors for join/group/having scope with certain values (0 and 1)
Testing ransack v1.6.3, rails 4.1.8

```ruby
class Cart < ActiveRecord::Base
  has_many :items

  def self.items_count_equals(n)
    joins(:items).group("carts.id").having("count(items.id) = ?", n)
  end

  def self.ransackable_scopes
    [:items_count_equals]
  end
end

# when 0 is passed in to ransack, no filtering is performed, and all Cart records are
# returned. (in fact, the scope method is never called)
# meanwhile the class method *does* produce correct results with 0 passed in
> Cart.ransack(items_count_equals: 0).result.length == Provider.items_count_equals(0).length
=> false

# when 1 is passed in to ransack, the scope method is called
# but `true` comes in for `n` which is eventual converted to
# a "t", which causes a SQL error
> Cart.ransack(items_count_equals: 1).result
PG::InvalidTextRepresentation: ERROR:  invalid input syntax for integer: "t"
LINE 1: ...rts"."id" GROUP BY carts.id HAVING count(items.id) = 't'
```

Passing in positive integer values other than 0 and 1 appears to produce the expected results.

This unexpected behavior also appears when using named scopes rather than class methods and when querying for `<` and `>` in addition to `=`.

When using named a named scope and a lambda, the error is different in the case when 1 is passed in:
```ruby
ArgumentError: wrong number of arguments (0 for 1)
```

This suggests to me this bug may be connected to #404 .

gkop:
The TRUE_VALUES and FALSE_VALUES haven't been changed since the initial commit 294015309b03240c6a9a9aea69118a24d7f737ac , yet my removing the values that I have removed doesn't break any tests.

@ernie do you remember the purpose for the values I propose removing?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+*   When sanitizing scope arguments, do not replace 0 and 1 with false and true
+    cf. https://github.com/activerecord-hackery/ransack/issues/502
+
+    *Gabe Kopley*
+
+
 ## Version 1.6.3 - 2015-01-21
 
 *   Fix a regression
--- a/lib/ransack/constants.rb
+++ b/lib/ransack/constants.rb
@@ -40,8 +40,8 @@ module Ransack
     STASHED_JOIN        = 'stashed_join'.freeze
     JOIN_NODE           = 'join_node'.freeze
 
-    TRUE_VALUES = [true, 1, '1', 't', 'T', 'true', 'TRUE'].to_set
-    FALSE_VALUES = [false, 0, '0', 'f', 'F', 'false', 'FALSE'].to_set
+    TRUE_VALUES = [true, 't', 'T', 'true', 'TRUE'].to_set
+    FALSE_VALUES = [false, 'f', 'F', 'false', 'FALSE'].to_set
     BOOLEAN_VALUES = (TRUE_VALUES + FALSE_VALUES).freeze
 
     S_SORTS             = %w(s sorts).freeze
--- a/spec/ransack/adapters/active_record/base_spec.rb
+++ b/spec/ransack/adapters/active_record/base_spec.rb
@@ -20,7 +20,8 @@ module Ransack
 
           context 'with scopes' do
             before do
-              Person.stub :ransackable_scopes => [:active, :over_age, :of_age]
+              Person.stub :ransackable_scopes =>
+                [:active, :over_age, :of_age, :article_count_equals]
             end
 
             it "applies true scopes" do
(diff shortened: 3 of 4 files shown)


gkop:
Travis fails this on Rails 3.0, but it's not my code change causing the failure, but the additional tests I have added...

jonatack:
Thanks @gkop. If memory serves, Rails 3.0 and 3.1 are no longer supported by the Rails team and are basically at end of life status. Let's try merging this in, and add a conditional to not run these tests for 3.0.

We are getting pretty close to the time when Rails 3.0 and 3.1 support could probably be removed from Ransack master and from Ransack 2.x. Anyone who is running Rails 3 could use Ransack up to 1.6.x.

avit:
The reason for those values is because rails form checkboxes post "1" or "0" to mean true or false, which was adapted for scopes to mean "apply this scope (yes/no)", *not* as a scope argument.

Passing arguments to scopes was designed to be done as an array; this change will break things. I think #502 is a documentation/usage issue rather than a bug.

```
# before:
Person.ransack("admin" => "1")          # calls Person.admin
Person.ransack("over_age" => ["18"])    # calls Person.over_age(18)

# after:
Person.ransack("admin" => "1")          # calls Person.admin(1) #boom!
Person.ransack("over_age" => ["18"])    # calls Person.over_age(18)
```

Note that scopes can take multiple arguments, so an array is necessary. Passing a single argument without the array was just a convenience for values other than "1" and "0" from checkboxes.

@jaredbeck that's what you're talking about right?

jonatack:
Thanks @avit, you're right :+1:

jaredbeck:
> @jaredbeck that's what you're talking about right?

Thanks for the reply.  Yeah, I think so.

So if `my_bool` is a boolean column in the database, and there's a checkbox like this:

    <input name="search[my_bool_eq]" value="1">

that goes straight into a search, like:

    MyRecord.search(params[:search])

You're saying that's *not* an intended use of ransack?  I should transform `params[:search]` into a hash like `my_bool: true`?  Note that I'm not talking about an AR scope, just a boolean column in a db table.

jonatack:
Will revert. We need a test that covers this.

jaredbeck:
> We need a test that covers this.

Please check out https://github.com/activerecord-hackery/ransack/pull/507 for some tests to get us started.  They pass in 1.6.3, and fail after https://github.com/activerecord-hackery/ransack/pull/503

jonatack:
Thank you, @jaredbeck! My initial thought is to prefer adding these tests in the `predicate_spec.rb` suite and test for all the values. Not sure how compatible this is with MySQL, though. Let's see what Travis-ci says.

Issue #504, opened 2015-01-30
1.6.0 works, but master has NoMethodError: undefined method 'val'


jaredbeck:
NoMethodError: undefined method `val' for 
    #<Arel::Nodes::Quoted:0x007fd3740118b0 @expr=nil>

This issue seems to have been introduced in 1.6.1 by https://github.com/activerecord-hackery/ransack/commit/6c59e96adaa43f4e8141352f508b84a046343591

If I *revert* the following change to `adapters/active_record/ransack/nodes/condition.rb` from said commit, then my tests pass again.

```diff
-          predicate.right.is_a?(Array) &&
-          predicate.right[0].class == Arel::Nodes::Casted
+          predicate.right[0].val.is_a?(Array)
```

Of course, I'm not suggesting we revert the change, just trying to help isolate the issue.

Stack trace:

```
NoMethodError: undefined method `val' for #<Arel::Nodes::Quoted:0x007fc1d4872678 @expr=nil>
..snip..lib/ransack/adapters/active_record/ransack/nodes/condition.rb:48:in `casted_array_with_in_predicate?'
..snip..lib/ransack/adapters/active_record/ransack/nodes/condition.rb:37:in `block in format_predicate'
..snip..lib/ransack/adapters/active_record/ransack/nodes/condition.rb:36:in `tap'
..snip..lib/ransack/adapters/active_record/ransack/nodes/condition.rb:36:in `format_predicate'
..snip..lib/ransack/adapters/active_record/ransack/nodes/condition.rb:23:in `arel_predicate_for'
..snip..lib/ransack/adapters/active_record/ransack/nodes/condition.rb:6:in `arel_predicate'
..snip..lib/ransack/visitor.rb:17:in `visit_Ransack_Nodes_Condition'
..snip..lib/ransack/visitor.rb:52:in `visit'
..snip..lib/ransack/visitor.rb:5:in `accept'
..snip..lib/ransack/adapters/active_record/ransack/visitor.rb:4:in `block in visit_and'
..snip..lib/ransack/adapters/active_record/ransack/visitor.rb:4:in `map'
..snip..lib/ransack/adapters/active_record/ransack/visitor.rb:4:in `visit_and'
..snip..lib/ransack/visitor.rb:24:in `visit_Ransack_Nodes_Grouping'
..snip..lib/ransack/visitor.rb:52:in `visit'
..snip..lib/ransack/visitor.rb:5:in `accept'
[...]

PR #505, opened 2015-01-30
Fix NoMethodError in casted_array_with_in_predicate?

jaredbeck:
This failing test reproduces https://github.com/activerecord-hackery/ransack/issues/504

It's a much-simplified version of a ransacker in my application which worked in 1.5.1 and 1.6.0, but which fails in subsequent releases, including master, because of https://github.com/activerecord-hackery/ransack/commit/6c59e96adaa43f4e8141352f508b84a046343591 as explained in https://github.com/activerecord-hackery/ransack/issues/504

Disclaimer: This is the first time I've ever worked with ransack or arel in detail, and I'm not the one who wrote the ransacker in my app.  I'm just trying to upgrade to the latest ransack (1.6.3).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .bundle
 Gemfile.lock
 pkg/*
+.idea
--- a/lib/ransack/adapters/active_record/ransack/nodes/condition.rb
+++ b/lib/ransack/adapters/active_record/ransack/nodes/condition.rb
@@ -43,6 +43,7 @@ module Ransack
         def casted_array_with_in_predicate?(predicate)
           return unless defined?(Arel::Nodes::Casted)
           predicate.class == Arel::Nodes::In &&
+          predicate.right[0].respond_to?(:val) &&
           predicate.right[0].val.is_a?(Array)
         end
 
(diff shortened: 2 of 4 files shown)


jaredbeck:
@jonatack please let me know how you'd like to proceed with this.

jaredbeck:
> let me know how you'd like to proceed

What about a simple solution like `respond_to?(:val)`?

```diff
  def casted_array_with_in_predicate?(predicate)
    return unless defined?(Arel::Nodes::Casted)
    predicate.class == Arel::Nodes::In &&
+     predicate.right[0].respond_to?(:val) &&
      predicate.right[0].val.is_a?(Array)
  end
```

jaredbeck:
If I understand this all correctly, the right-hand side of an `IN` predicate does not always respond to :val.  For example, it might be a subquery, represented by a `SelectManager`.

jonatack:
Thank you, this was helpful in reproducing the issue and I've written a failing/passing spec based on what you've shown above.

OK for the quick fix with `respond_to?` as above or using `try` like `predicate.right[0].try(:val).is_a?(Array)`.

When I get some time I need to dig in and go through the new Arel and AR API changes. There may be a better way of handling these cases.

jaredbeck:
> Thank you, this was helpful 

No problem.  Thank you for being so quick to respond throughout the day!

Like I said, this was my first time working with ransack and arel in any detail and I learned a ton.

PR #506, opened 2015-02-01
Gitignore rubymine's .idea directory

jaredbeck:
If everyone on the project used Intellij / rubymine, we might want to commit the `.idea` directory some day, but, as this is OSS and it's unlikely that everyone will use the same IDE, we'll probably never want to commit `.idea`.  So, git-ignoring it is a matter of convenience for rubymine users.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .bundle
 Gemfile.lock
 pkg/*
+.idea


jonatack:
Correct me if I'm wrong, but I would think this kind of personal configuration would best be located in a developer's global gitignore file.

jaredbeck:
> .. this kind of personal configuration would best be located in a developer's global gitignore file.

I didn't know such a thing existed.  Worked for me.  Thanks!

PR #507, opened 2015-02-01
Add tests of how boolean columns worked in 1.6.3

jaredbeck:
The string '1', in 1.6.3, was translated into boolean true.

These tests pass in 1.6.3 and fail after 9f4e48e.

CC: https://github.com/activerecord-hackery/ransack/pull/503

### DIFF
--- a/spec/ransack/adapters/active_record/base_spec.rb
+++ b/spec/ransack/adapters/active_record/base_spec.rb
@@ -144,6 +144,25 @@ module Ransack
             s = Person.ransack('')
           end
 
+          context 'boolean column' do
+            let(:is_mysql) { ENV["DB"] == "mysql" }
+            let(:sql_col_ref) { is_mysql ? %Q{`people`.`awesome`} : %Q{"people"."awesome"} }
+
+            it "should find true records" do
+              sql_true = is_mysql ? '1' : "'t'"
+              s = Person.ransack(:awesome_eq => '1')
+              expected = "%s = %s" % [sql_col_ref, sql_true]
+              expect(s.result.to_sql).to include(expected)
+            end
+
+            it "should find false records" do
+              sql_false = is_mysql ? '0' : "'f'"
+              s = Person.ransack(:awesome_eq => '0')
+              expected = "%s = %s" % [sql_col_ref, sql_false]
+              expect(s.result.to_sql).to include(expected)
+            end
+          end
+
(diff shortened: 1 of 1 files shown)


jaredbeck:
Hmm looks like mysql uses a `1` for boolean true, while others use a char `'t'`.  So, the mysql tests are failing.  Oh well, you get the idea. :)

jaredbeck:
Maybe the test could switch on `ENV['DB']`, or `ActiveRecord::Base.connection.adapter_name`?

Or we could drop the right-hand side of the SQL comparison in our test assertion.  Just testing the presence of the column would be an improvement over `master`.

avit:
Suggest: `sql_true = Person.connection.quote(true)`

jonatack:
OK, just saw this, thanks :+1: My initial thought is to prefer adding these tests in the predicate_spec.rb suite and test for all the values. Let's see what Travis-ci says about it.

jaredbeck:
> Suggest: sql_true = Person.connection.quote(true)

Nice.  Thanks!  Is there a quoting method for `table.column`?

jonatack:
Yes:

```ruby
def expected_boolean_query(boolean_value)
  field = "#{quote_table_name("people")}.#{quote_column_name("awesome")}"
  condition = ActiveRecord::Base.connection.send("quoted_#{boolean_value}")
  /#{field} = #{condition}/
end
```

jaredbeck:
Cool, I'll give `quote_column_name` a try.  I was unsure, because I didn't see an implementation of it in e.g. https://github.com/rails/rails/blob/3-2-stable/activerecord/lib/active_record/connection_adapters/mysql2_adapter.rb

avit:
Thanks guys! :)

jonatack:
Thanks to both of you for the heads up!

jaredbeck:
> My initial thought is to prefer adding these tests in the predicate_spec.rb suite and test for all the values.

Makes sense, especially the "test for all the values" part.  My tests here only cover char '0' and '1'.  Thanks!

jonatack:
@avit thanks for the suggestion. In my last commit, `condition = ActiveRecord::Base.connection.quote(boolean_value)` would be an improvement over `condition = ActiveRecord::Base.connection.send("quoted_#{boolean_value}")`, if you'd like to change it.

PR #509, opened 2015-02-02
Failing test cases for when 0 and 1 are passed to non-boolean scopes

gkop:
@avit, may we re-open #504 based on these failing test cases?

### DIFF
--- a/spec/ransack/adapters/active_record/base_spec.rb
+++ b/spec/ransack/adapters/active_record/base_spec.rb
@@ -20,8 +20,9 @@ module Ransack
 
           context 'with scopes' do
             before do
-              Person.stub :ransackable_scopes => [:active, :over_age, :of_age]
-            end
+              Person.stub :ransackable_scopes =>
+                [:active, :over_age, :of_age, :article_count_equals]
+             end
 
             it "applies true scopes" do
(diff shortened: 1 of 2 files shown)


avit:
It looks like this is passing except that your tests expect `HAVING count(articles.id) = 0` but the actual SQL query is quoted for different databases as `HAVING count(articles.id) = 'f'`

See #507

avit:
Ah, never mind. I see this should be an integer count...

jaredbeck:
> @avit, may we re-open #504 based on these failing test cases?

@gkop you meant to say "re-open https://github.com/activerecord-hackery/ransack/pull/503", right?

gkop:
Derp, yes you got it Jared!

gkop:
Actually @jaredbeck I meant #502 !

avit:
Looking at this again after a long while... 

I don't think Ransack should be responsible for converting params anymore, this should probably be happening in the controller anyway. We should probably just support this without getting too clever:

* Array for params passed as arguments, without converting values
* Single value for selecting the scope (truthy/falsy), without arguments

seanfcarroll:
Superseded by https://github.com/activerecord-hackery/ransack/pull/742

Issue #508, opened 2015-02-02
FR: Ransack::Condition is accessible in Ransack::Attribute


sineed:
Hi! I have a feature request.

Sometimes it can be useful to have instance of `Ransack::Condition` as instance variable of `Ransack::Attribute`. If so, condition can be bound into ransacker block. You can create a ransacker that can be based on condition values. For example,

```ruby
ransacker :test, args: [:condition] do |condition|
  value = condition.values.first.value
  if some_condition_based_on(value)
    Arel.sql('some SQL')
  else
    Arel.sql('another SQL')
  end
end
```

I don't know if it fits to main philosophy of `ransack` so before implementing some code I want to know if it can be useful not only for me.

avit:
I do some funky things in ransackers to work around things myself, but I don't remember off the top of my head. Can you give an example use case? (the search params & the resulting query)

jonatack:
Thanks. It might be useful to people and I'm curious to see an example as well.

sineed:
@avit @jonatack Thanks for the quick feedback.

In my current project there are entities (say, `Product`) which can have custom fields  (say, `CustomField`). One `Product` has one `CustomFieldValue` also `CustomFieldValue` has a reference to `CustomField`:

```ruby
[21] pry(main)> CustomFieldValue
=> CustomFieldValue(id: integer, value: string, product_id: integer, custom_field_id: integer, created_at: datetime, updated_at: datetime) 
```

I want to search products by custom field value that has specific custom field id. And I want to use predicate logic as well for `CustomFieldValue`' value field.

So If `Ransack::Attribute` has a reference to its `Ransack::Condition` I need only *one* ransacker:

```ruby
ransacker :custom_field_value, args: [:condition] do |condition|
  custom_field_id = condition.values.shift
  Arel.sql("(SELECT value FROM custom_field_values cfv INNER JOIN custom_fields cf ON cfv.custom_field_id = cf.id WHERE products.id = cfv.product_id AND cf.id = #{custom_field_id.value})")
end
```

And usage of this ransacker:

```ruby
Product.ransack(c: [{ v: [custom_field_id, value], p: 'eq', a: ["custom_field_value"] }])
```

I don't like that I pass custom_field_id in `values` key. Maybe it can be another key in `Condition` build options that can be accessed in `Attribute` instance (e.g. `meta`)?

@jonatack is this still useful for other people?

avit:
[...]

PR #513, opened 2015-02-18
Allow ransacker to have its arguments

sineed:
This is related to https://github.com/activerecord-hackery/ransack/issues/508 but I have another use case:

In my project I have Customer and Deal models. Customers have many Deals and each deal has `amount` attribute. I want to find customers which have made deals with total amount more than specified value for the last week. So my ransacker is:

```ruby
# app/models/customer.rb
ransacker :sum_of_deals_last_week do |parent|
  deals_table = Deal.arel_table

  Arel::Nodes::Grouping.new(
    deals_table
      .project(Arel::Nodes::Sum.new([deals_table[:amount]]))
      .group(deals_table[:customer_id])
      .where(deals_table[:customer_id].eq(parent.table[:id])
        .and(deals_table[:created_at].gteq(1.week.ago.to_date)
          .and(deals_table[:created_at].lteq(Date.today))
        )
      ).ast
  )
end
```

And I can easily filter:

```ruby
Customer.ransack(sum_of_deals_last_week_gt: 100)
```

The problem occurs when I want to change period of time. Say, for last month period I need to create new ransacker. What if I need to dynamically specify this period?

So with changes in this PR the following ransacker can be created:

```ruby
# app/models/customer.rb
ransacker :sum_of_deals, args: [:parent, :rargs] do |parent, rargs|
  date_from = rargs[0]
  date_till = rargs[1]
  deals_table = Deal.arel_table

  Arel::Nodes::Grouping.new(
    deals_table
      .project(Arel::Nodes::Sum.new([deals_table[:amount]]))
      .group(deals_table[:customer_id])
      .where(deals_table[:customer_id].eq(parent.table[:id])
        .and(deals_table[:created_at].gteq(date_from)
          .and(deals_table[:created_at].lteq(date_till))
        )
      ).ast
  )
end
```

And I can use this ransacker:

```ruby
Customer.ransack(
  c: [{
    a: { '0' => { name: :sum_of_deals, rargs: [1.week.ago.to_date, Date.today] } },
    p: :gt,
    v: [100]
  }]
)
```

### DIFF
--- a/lib/ransack/nodes/attribute.rb
+++ b/lib/ransack/nodes/attribute.rb
@@ -3,14 +3,15 @@ module Ransack
     class Attribute < Node
       include Bindable
 
-      attr_reader :name
+      attr_reader :name, :rargs
 
       delegate :blank?, :present?, :==, :to => :name
       delegate :engine, :to => :context
 
-      def initialize(context, name = nil)
+      def initialize(context, name = nil, rargs = [])
         super(context)
         self.name = name unless name.blank?
+        @rargs = rargs
       end
 
       def name=(name)
--- a/lib/ransack/nodes/condition.rb
+++ b/lib/ransack/nodes/condition.rb
@@ -66,7 +66,7 @@ module Ransack
           end
         when Hash
           args.each do |index, attrs|
-            attr = Attribute.new(@context, attrs[:name])
+            attr = Attribute.new(@context, attrs[:name], attrs[:rargs])
             self.attributes << attr if attr.valid?
           end
         else
--- a/spec/ransack/adapters/active_record/base_spec.rb
+++ b/spec/ransack/adapters/active_record/base_spec.rb
@@ -314,7 +314,7 @@ module Ransack
             )
           end
 
-          it 'searches by ransacker with ransacker_attributes', focus: true do
+          it 'searches by ransacker with ransacker_attributes' do
             s = Person.ransack(
               c: [{
                 a: {
@@ -324,7 +324,7 @@ module Ransack
                   }
                 },
                 p: 'cont',
-                v: ['Rails 5.0 released']
+                v: ['Rails has been released']
               }]
             )
 
@@ -333,7 +333,7 @@ module Ransack
             )
 
             expect(s.result.to_sql).to match(
-              /LIKE \'\%Rails 5\.0 released\%\'/
+              /LIKE \'\%Rails has been released\%\'/
             )
           end
         end
--- a/spec/ransack/adapters/active_record/base_spec.rb
+++ b/spec/ransack/adapters/active_record/base_spec.rb
@@ -313,6 +313,29 @@ module Ransack
                 quote_column_name("only_admin")} = 'htimS cirA'/
             )
           end
+
+          it 'searches by ransacker with ransacker_attributes', focus: true do
+            s = Person.ransack(
+              c: [{
+                a: {
+                  '0' => {
+                    name: 'articles_title_with_specified_body_size',
+                    rargs: [10, 100]
+                  }
+                },
+                p: 'cont',
+                v: ['Rails 5.0 released']
+              }]
+            )
+
+            expect(s.result.to_sql).to match(
+              /CHAR_LENGTH\(articles.body\) BETWEEN 10 AND 100/
+            )
+
+            expect(s.result.to_sql).to match(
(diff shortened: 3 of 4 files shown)


jonatack:
Thank you, @sineed. Would you like to update the wiki about this API?

I'll admit I'm hesitant about the name `rargs`. How about `r_args` or `ransacker_args`?

sineed:
@jonatack I've updated wiki page _Using Ransackers_ and added #519

jonatack:
@sineed thanks. The test and wiki example query generates the following SQL query, which does not seem to be valid SQL:

```
SELECT "people".* FROM "people" WHERE ((SELECT title FROM articles WHERE articles.person_id = people.id AND CHAR_LENGTH(articles.body) BETWEEN 10 AND 100 ) LIKE '%Rails has been released%')  ORDER BY "people"."id" DESC
```

Could you create an example that generates valid SQL?

railsdevmatt:
@sineed @jonatack How would this be used in a `sort_link` I have a sort link like the following
```ruby
= sort_link @q, :read_topics, 'Only Items Ive read'
```
which I want to pass over the current_user id, is that possible with this? The wiki only does not make it clear on how to pass an argument like this

sineed:
@jonatack 
I've updated #519 and wiki. Please take a look.

@railsdevmatt 
I don't use sorting functionality of `ransack`. And I don't know if a ransacker can be used as an attribute for `sort_link` method.
Ransacker arguments were implemented primary for advanced searching. And maybe it's possible to perform sorting without need of ransacker, can you add more info about your use case?

railsdevmatt:
Ok so i have a forum with topics
```ruby
# forums controller

def show
  @forum = Forum.find(params[:id])
  @q = @forum.topics.ransack(params[:q])
  @topics = @q.result.page(params[:page])
end

# view show.html.haml

- unless @topics.empty?

  %ul
    %li= sort_link @q, :read_topics, 'Only topics Ive read'

  - @topics.each do |topic|
    ...
    ...
```
I want to pass in the current_user to the read_topics ransacker that i created and filter only the read topics. I do know this is possible without ransack, but figured if there was a way to accomplish this with a ransacker then I would give it a go.

sineed:
@railsdevmatt Am I right if your topics have a `have_and_belongs_to_many` association with current_user' class? If so you can use https://github.com/activerecord-hackery/ransack#associations logic, something like this (say your association is named `readers`):

```ruby
  link_to 'Only topics Ive read', forum_path(id: @forum.id, q: { readers_user_id: current_user.id })
```

railsdevmatt:
@sineed - actually the relationship is like this

```ruby
# topic model
class Topic < ActiveRecord::Base
  has_many :topic_reads, dependent: :destroy
  has_many :subscribers, through: :topic_subscriptions
```
```ruby
class TopicRead < ActiveRecord::Base
  belongs_to :topic
  belongs_to :reader, class_name: "User", foreign_key: "reader_id"
```
```ruby
class TopicSubscription < ActiveRecord::Base
  belongs_to :topic
  belongs_to :subscriber, -> { where "users.email != '' AND users.email IS NOT NULL" }, class_name: "Contact", foreign_key: "subscriber_id"
```

jonatack:
The sort_link is for setting the `order_by` SQL via the Ransack `:sorts` (`:s`) option.

Any filtering should be happening in your controller action (where you have ready access to `current_user`), either based on if the sort is set to a particular value, or in general.

reubenbrown:
@sineed @jonatack Love this feature!  If I were constructing a ransack query in my controller like so (by "manually" merging together attribute names with predicates):
```
User.ransack({"s"=>"created_at desc", "created_at_eq"=>"2015-06-10"}).result
```
And I wanted to pass arguments to my ransacker for `created_at`
```
ransacker :created_at, args: [:parent, :ransacker_args] do |parent, args|
  start_time, end_time = args
  User.where('users.created_at > ? AND users.created_at < ?', start_time, end_time)
end
```
Is it possible to pass `ransacker_args` given the way I am constructing my query hash?

I've been playing around trying to properly include it in my hash but keep getting an undefined method error ransacker_args for the sort node. Since the [documentation](https://github.com/activerecord-hackery/ransack/wiki/Using-Ransackers) provides only one example of passing arguments to a ransacker it's a bit unclear if what I'm trying to do is even possible.

PR #514, opened 2015-02-18
Allow to use model name from I18n ransack scope

nagyt234:
This simple pull request allow us to use I18n model names from ransack scope, not only from activerecord scope, similarly as for attribute names.
Background:
1. I need in attribute select different name for the base model as generally.
2. So is the using of I18n in ransack unified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -632,6 +632,8 @@ en:
       end: ends with
       gt: greater than
       lt: less than
+    models:
+      person: Passanger
     attributes:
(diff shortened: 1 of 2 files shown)


jonatack:
@nagyt234 Thank you!

jonatack:
Could you add a test, please?

jonatack:
@nagyt234 Thank you! Could you add a test?

PR #526, opened 2015-03-28
Fix erroneous reference to AR class

ncuesta:
9e818cde introduced a bug by making an erroneous reference to `ActiveRecord::Associations::JoinDependency` in line 187. As this class lives inside an `ActiveRecord` module, that reference needs to be absolute in order to properly point at AR's class.

### DIFF
--- a/lib/ransack/adapters/active_record/context.rb
+++ b/lib/ransack/adapters/active_record/context.rb
@@ -184,7 +184,7 @@ module Ransack
               :string_join
             when Hash, Symbol, Array
               :association_join
-            when ActiveRecord::Associations::JoinDependency
+            when ::ActiveRecord::Associations::JoinDependency
               :stashed_join
             when Arel::Nodes::Join
               :join_node
--- a/lib/ransack/adapters/active_record/context.rb
+++ b/lib/ransack/adapters/active_record/context.rb
@@ -184,7 +184,7 @@ module Ransack
               :string_join
             when Hash, Symbol, Array
               :association_join
-            when ::ActiveRecord::Associations::JoinDependency
+            when Polyamorous::JoinDependency, Polyamorous::JoinAssociation
               :stashed_join
             when Arel::Nodes::Join
               :join_node


jonatack:
@ncuesta Thank you for the pull request. Can you say more about what issue this causes and how it fails (stack trace)? Is there a test we can add? I just ran the tests with both versions, as well as using the `JoinDependency, JoinDependency::JoinAssociation` for the conditional, and the test suite passes with all three versions :mag:

jonatack:
@ncuesta Thank you for the pull request. I'm not up to speed with AR namespacing. Could you tell me more about what issue this causes :bomb:? Stack trace? Is there a test we can add? I ran the tests before/after your PR, as well as using the `JoinDependency, JoinDependency::JoinAssociation`, and the test suite passes with all three versions.

jonatack:
@ncuesta Thanks for the heads up. After looking at this, it might be best to use the `JoinDependency, JoinAssociation` defined in Polyamorous [here](https://github.com/activerecord-hackery/polyamorous/blob/master/lib/polyamorous.rb). What do you think? If you agree, could you please update this PR and we'll merge it in. Thanks again.

ncuesta:
@jonatack Even better! That way this references from ransack to external classes won't be tied to specific versions of AR/Arel. I'll update the PR in a moment.

ncuesta:
@jonatack Done, tested it locally on my project and things are working just fine. Great insight on Polyamorous, hadn't checked that *lovely* gem :)

jonatack:
Thank you @ncuesta.

This has made me realize that there may some optimizing and de-duplicating that could be done. These values are defined both in the various AR versions of Ransack::Context and also in Polyamorous.

BTW if you are good at meta-programming and feel like adding more commits, the `alias_method_chain` methods in Polyamorous in the Rails 5.0 versions of the join_dependency and join_association files need to be converted to `prepend` with `super`... `alias_method_chain` has been deprecated in Rails 5.0 ;)

ncuesta:
@jonatack I'll try and help as soon as I get some room for it. Glad to be able to help!

Issue #253, opened 2013-07-15
Short alias for long compound search field


estum:
I have search field with very long parameter name: `user_bill_info_fullname_or_user_bill_info_phone_or_user_bill_info_address_cont`. How to create short alias for it, eg `bill_info`?

jonatack:
Added with #623.

PR #536, opened 2015-05-03
Create aliases for (long) search fields

glampr:
Allow to alias (possibly long) search fields with (shorter) descriptions.
Works like ransacker, but uses already valid search fields, instead of requiring to write Arel.
Useful for converting long search fields to shorter, to save url length for GET requests.

Usage
```
class User
  translates :profile_text # from globalize gem
  has_many :works

  ransacker_alias :free_text, :translations_profile_text_or_works_translations_title_or_works_translations_description
end

User.search(free_text_cont: "some_text")
# same as User.search(translations_profile_text_or_works_translations_title_or_works_translations_description_cont: "some_text")
```

### DIFF
--- a/lib/ransack/adapters/active_record/base.rb
+++ b/lib/ransack/adapters/active_record/base.rb
@@ -14,6 +14,14 @@ module Ransack
         end
 
         def ransack(params = {}, options = {})
+          params.keys.each do |k|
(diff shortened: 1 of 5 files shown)


glampr:
Also mentioned in #253

jonatack:
Thanks for the contribution. Closed by #623.

PR #541, opened 2015-05-22
ActiveRecord excludes Mongoid

ASnow:
Adapters overwrite each other, so they should not be required both

### DIFF
--- a/lib/ransack.rb
+++ b/lib/ransack.rb
@@ -2,11 +2,8 @@ require 'active_support/core_ext'
 
 require 'ransack/configuration'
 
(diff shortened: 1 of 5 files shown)


jonatack:
@ASnow LGTM. I'm sorry to have taken so long to review your PR. Good idea to extract the logic all over the codebase into one place. Travis-ci not passing put me off, but it appears to be a non-critical failure, so let's merge this in and make it work.

jonatack:
Thanks again for this PR :+1:

jonatack:
Fixes #549.

PR #560, opened 2015-07-08
Avoid overwriting association conditions with default scope in Rails 3

avit:
This is a fix on top of my previous pull request #426 where I fixed one bug but potentially introduced another. :speak_no_evil: 

When a model with default scope was associated with conditions (`has_many :x, conditions: ...`), the default scope would overwrite the association conditions. This patch ensures that both sources of conditions are applied.


avit:
Found issues: it doesn't handle conditions on aliased table names, and I think we can avoid altering the conditions on the original association reflection (this is held by the parent class globally so it's best not to mutate anything on it). I will resubmit with some improvements.

jonatack:
Thanks @avit!
:smiley:

jonatack:
Could you update the change log with your fix?

PR #561, opened 2015-07-09
Fix time-dependent test failure

avit:
When the database has `default_timezone = :local` (system time) and the Time.zone is set to elsewhere then `Date.current` does not match what the query produces for the stored timestamps.

Resolved by setting everything to UTC.

### DIFF
--- a/spec/ransack/adapters/active_record/base_spec.rb
+++ b/spec/ransack/adapters/active_record/base_spec.rb
@@ -145,6 +145,9 @@ module Ransack
           end
 
           it "should function correctly with a multi-parameter attribute" do
+            ::ActiveRecord::Base.default_timezone = :utc
+            Time.zone = 'UTC'
+
             date = Date.current
             s = Person.ransack(
               { "created_at_gteq(1i)" => date.year,


jonatack:
Thanks @avit!

jonatack:
Could you update the change log with your fix?

spec/ransack/adapters/active_record/base_spec.rb
```diff
@@ -145,6 +145,9 @@ def self.sane_adapter?
           end
 
           it "should function correctly with a multi-parameter attribute" do
+            ::ActiveRecord::Base.default_timezone = :utc
```
nj: Would have been better to have this in `config/environments/test.rb`, as this could lead to unexpected behaviour for other specs.

spec/ransack/adapters/active_record/base_spec.rb
```diff
@@ -145,6 +145,9 @@ def self.sane_adapter?
...
+            Time.zone = 'UTC'
```
nj: Normally, if just temporary, a `Time.use_zone` block around the related code might have been better - here also to avoid unexpected behaviour other places in the test suite.

spec/ransack/adapters/active_record/base_spec.rb
jonatack: Hm... a different TZ is set in spec_helper.rb as well. Suggestion?

spec/ransack/adapters/active_record/base_spec.rb
jonatack: Would you like to make a PR?

PR #562, opened 2015-07-12
Fix handling attribute-name with `_and_` `_or_`

### LINKED ISSUE #299: Ransack doesn't handle attribute names including "and" / "or"

hoshitocat:
This PR tries to resolve #299 

I think Ransack must recognize attribute\_name which includes `_and_` `_or_`.

My approach is:
If the model includes attribute\_name, Ransack should recognize it as one attribute\_name.

Please feedback this PR !!

### DIFF
--- a/lib/ransack/nodes/condition.rb
+++ b/lib/ransack/nodes/condition.rb
@@ -31,7 +31,7 @@ module Ransack
 
         private
 
-        def extract_attributes_and_predicate(key, context=nil)
+        def extract_attributes_and_predicate(key, context = nil)
           str = key.dup
           name = Predicate.detect_and_strip_from_string!(str)
           predicate = Predicate.named(name)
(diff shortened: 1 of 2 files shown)


jonatack:
Thanks @hoshitocat! Nice! Could you update the change log with this fix?

hoshitocat:
Thank you @jonatack !!
I'm happy to merge PR :)
I will udate CHANGELOG.md soon.

PR #564, opened 2015-07-14
not requiring mongoid if activerecord is defined. This is so it works…

miguelperez:
… by default with activerecord and activeadmin wont break.

This is because I have an application that have both activerecord and mongoid. The thing is that I am only listing activerecord models in my active admin, so there is no need for loading mongoid's ransack modules.


miguelperez:
sorry for this... I accidentally created this pr.

PR #576, opened 2015-08-19
note that AR and Mongoid don't work together

timoschilling:


jonatack:
Thank you! For documentation-only changes, please remember to add `[skip ci]` to the commit message so travis-ci does not :runner: needlessly.

PR #577, opened 2015-08-21
add functionality to hide all search indicator arrows via config

kalami:
config/initializer/ransack.rb

Ransack.configure do |c|
  c.remove_search_order_indicators = true
end

This is basically like passing 'hide_indicator: true' to all of your search indicators..without passing it to all of your search indicators.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,7 +218,7 @@ by adding this to an initializer like `config/initializers/ransack.rb`:
 
 ```ruby
 Ransack.configure do |c|
-  c.remove_search_order_indicators = true
+  c.hide_sort_order_indicators = true
 end
 ```
 
--- a/README.md
+++ b/README.md
@@ -213,7 +213,8 @@ The sort link may be displayed without the order indicator arrow by passing
 <%= sort_link(@q, :name, hide_indicator: true) %>
 ```
 
-Alternatively, all sort links may be displayed without the order indicator arrow by adding this to an initializer like config/initializers/ransack.rb:
+Alternatively, all sort links may be displayed without the order indicator arrow 
+by adding this to an initializer like `config/initializers/ransack.rb`:
 
 ```ruby
 Ransack.configure do |c|
--- a/README.md
+++ b/README.md
@@ -213,6 +213,14 @@ The sort link may be displayed without the order indicator arrow by passing
 <%= sort_link(@q, :name, hide_indicator: true) %>
 ```
 
+Alternatively, all sort links may be displayed without the order indicator arrow by adding this to an initializer like config/initializers/ransack.rb:
+
+```ruby
+Ransack.configure do |c|
+  c.remove_search_order_indicators = true
+end
+```
+
 ### Advanced Mode
 
 "Advanced" searches (ab)use Rails' nested attributes functionality in order to
--- a/lib/ransack/configuration.rb
+++ b/lib/ransack/configuration.rb
@@ -9,7 +9,7 @@ module Ransack
     self.options = {
       :search_key => :q,
       :ignore_unknown_conditions => true,
-      :remove_search_order_indicators => false
+      :hide_sort_order_indicators => false
     }
 
     def configure
@@ -76,7 +76,7 @@ module Ransack
       end
     end
 
-    def remove_search_order_indicators=(boolean)
+    def hide_sort_order_indicators=(boolean)
       self.options[:remove_search_order_indicators] = boolean
     end
 
--- a/lib/ransack/configuration.rb
+++ b/lib/ransack/configuration.rb
@@ -77,7 +77,7 @@ module Ransack
     end
 
     def hide_sort_order_indicators=(boolean)
-      self.options[:remove_search_order_indicators] = boolean
+      self.options[:hide_sort_order_indicators] = boolean
     end
 
   end
--- a/lib/ransack/configuration.rb
+++ b/lib/ransack/configuration.rb
@@ -8,7 +8,8 @@ module Ransack
     self.predicates = {}
     self.options = {
       :search_key => :q,
-      :ignore_unknown_conditions => true
+      :ignore_unknown_conditions => true,
+      :remove_search_order_indicators => false
     }
 
     def configure
@@ -73,6 +74,10 @@ module Ransack
       else
         "#{arel_predicate}#{suffix}"
       end
+    end
+
+    def remove_search_order_indicators=(boolean)
+      self.options[:remove_search_order_indicators] = boolean
     end
 
   end
--- a/lib/ransack/helpers/form_helper.rb
+++ b/lib/ransack/helpers/form_helper.rb
@@ -94,7 +94,7 @@ module Ransack
           @current_dir    = existing_sort_direction
           @label_text     = extract_label_and_mutate_args!(args)
           @options        = extract_options_and_mutate_args!(args)
-          @hide_indicator = indicator?
+          @hide_indicator = @options.delete :hide_indicator || Ransack.options[:hide_sort_order_indicators]
           @default_order  = @options.delete :default_order
         end
 
@@ -197,16 +197,9 @@ module Ransack
             end
           end
 
-          def indicator?
-            @options.delete :hide_indicator unless Ransack.options[:hide_sort_order_indicators]
-          end
-
           def order_indicator
-            if @hide_indicator || no_sort_direction_specified? 
-              nil
-            else
-              direction_arrow
-            end
+            return if @hide_indicator || no_sort_direction_specified?
+            direction_arrow
           end
 
           def no_sort_direction_specified?(dir = @current_dir)
--- a/lib/ransack/helpers/form_helper.rb
+++ b/lib/ransack/helpers/form_helper.rb
@@ -198,7 +198,7 @@ module Ransack
           end
 
           def order_indicator
-            if @hide_indicator || no_sort_direction_specified?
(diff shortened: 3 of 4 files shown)


lib/ransack/helpers/form_helper.rb
```diff
@@ -198,7 +198,7 @@ def default_sort_order(attr_name)
           end
 
           def order_indicator
-            if @hide_indicator || no_sort_direction_specified?
+            if @hide_indicator || no_sort_direction_specified? || Ransack.options[:remove_search_order_indicators]
```
jonatack: Perhaps better to do this in #initialize where @hide_indicator is decided?

lib/ransack/configuration.rb
```diff
@@ -8,7 +8,8 @@ module Configuration
     self.predicates = {}
     self.options = {
       :search_key => :q,
-      :ignore_unknown_conditions => true
+      :ignore_unknown_conditions => true,
+      :remove_search_order_indicators => false
     }
```
jonatack: Perhaps call it `:remove_sort_order_indicators`, WDYT?

README.md
```diff
@@ -213,6 +213,13 @@ The sort link may be displayed without the order indicator arrow by passing
 <%= sort_link(@q, :name, hide_indicator: true) %>
 ```
 
+Alternatively, all sort links may be displayed without the order indicator arrow by adding this to an initializer like config/initializers/ransack.rb:
+
```
jonatack: Please wrap lines at 80 characters.

README.md
```diff
@@ -213,6 +213,13 @@ The sort link may be displayed without the order indicator arrow by passing
...
+    Ransack.configure do |c|
```
jonatack: For code formatting, use a block like:

```ruby
  code
```

README.md
```diff
@@ -213,6 +213,13 @@ The sort link may be displayed without the order indicator arrow by passing
...
+      c.remove_search_order_indicators = true
+    end
+
```
jonatack: Please use markdown ruby code format block here.

```ruby
  code
```

jonatack:
Thanks! Looks pretty good. I made a few comments in the code.

kalami:
You bet!  I'm glad you like it!

README.md
```diff
@@ -213,6 +213,15 @@ The sort link may be displayed without the order indicator arrow by passing
 <%= sort_link(@q, :name, hide_indicator: true) %>
 ```
 
+Alternatively, all sort links may be displayed without the order indicator arrow 
+by adding this to an initializer like `config/initializers/ransack.rb`:
+
+```ruby
+Ransack.configure do |c|
+  c.remove_search_order_indicators = true
```
jonatack: Perhaps should be updated to `c.hide_sort_order_indicators = true`?

lib/ransack/helpers/form_helper.rb
```diff
@@ -197,8 +197,12 @@ def default_sort_order(attr_name)
             end
           end
 
+          def indicator?
+            @options.delete :hide_indicator unless Ransack.options[:hide_sort_order_indicators]
+          end
+
           def order_indicator
-            if @hide_indicator || no_sort_direction_specified?
+            if @hide_indicator || no_sort_direction_specified?
```
jonatack: Can you remove the extra whitespace at the end of line 205?

We could shorten this to:

```ruby
def order_indicator
  return if @hide_indicator || no_sort_direction_specified?
  direction_arrow
end
```

lib/ransack/helpers/form_helper.rb
```diff
@@ -94,7 +94,7 @@ def initialize(search, attribute, args, params)
           @current_dir    = existing_sort_direction
           @label_text     = extract_label_and_mutate_args!(args)
           @options        = extract_options_and_mutate_args!(args)
-          @hide_indicator = @options.delete :hide_indicator
+          @hide_indicator = indicator?
```
jonatack: The `indicator?` method is a bit misnamed, it is really checking if Ransack should `hide_indicator?` and the logic in that method appears to be backward.

It might be better to inline this method in the constructor with this, WDYT?

```ruby
@hide_indicator = @options.delete(:hide_indicator) || Ransack.options[:hide_sort_order_indicators]
```

README.md
jonatack: Thanks for the better formatting!

spec/ransack/helpers/form_helper_spec.rb
```diff
@@ -399,6 +399,34 @@ module Helpers
...
+          Ransack.configure { |c| c.hide_sort_order_indicators = true }
+        end
+        subject { @controller.view_context
+          .sort_link(
+            [:main_app, Person.search(sorts: ['name desc'])],
+            :name,
+            controller: 'people'
+          )
+        }
+        it { should match /Full Name/ }
```
jonatack: 1. Does this test ever fail? Perhaps it should_not match `&nbsp;&#9660;`?

2. Could you please use RSpec `expect` syntax?

kalami: I think that if it is matching /Full Name/ it is automatically not matching either arrow.  I literally just copied/pasted that spec from the 'hide_indicator => true' spec because I figured the behavior is the same, but the implementation is different.

jonatack: I just tried that test with the inverse condition `hide_sort_order_indicators = false` and it still passes instead of failing.

lib/ransack/configuration.rb
```diff
@@ -75,5 +76,9 @@ def arel_predicate_with_suffix(arel_predicate, suffix)
       end
     end
 
+    def hide_sort_order_indicators=(boolean)
+      self.options[:remove_search_order_indicators] = boolean
```
jonatack: This was not updated to the new name yet.

spec/ransack/helpers/form_helper_spec.rb
jonatack: To make the test work properly, I'd suggest:
```ruby
it { should_not match /&#9660;|&#9650;/ }
```
Try it... it should pass when `hide_sort_order_indicators = true` and fail when `hide_sort_order_indicators = false`.

jonatack: OK for leaving the tests with RSpec `should` syntax since the rest of that file is that way. It might be better to move these 2 tests at the end of the `sort_link` tests where the other hide indicator specs are, and describe them like the other tests too, i.e. 'sort_link with...' instead of `search_form_for with...` since they relate to the sort_link method.

jonatack: If all this is confusing, don't worry about it and let me know, and I'll just commit it and clean it up.

kalami: Good catch, the spec for hide_indicator has the same problem.

lib/ransack/helpers/form_helper.rb
```diff
@@ -94,7 +94,7 @@ def initialize(search, attribute, args, params)
           @current_dir    = existing_sort_direction
           @label_text     = extract_label_and_mutate_args!(args)
           @options        = extract_options_and_mutate_args!(args)
-          @hide_indicator = @options.delete :hide_indicator
+          @hide_indicator = @options.delete :hide_indicator || Ransack.options[:hide_sort_order_indicators]
```
jonatack: `:hide_indicator` needs to be wrapped in parentheses here, otherwise it is evaluated by Ruby as `@options.delete (:hide_indicator || Ransack.options[:hide_sort_order_indicators])` which isn't what we want. It needs to be:

```ruby
          @hide_indicator = @options.delete(:hide_indicator) ||
            Ransack.options[:hide_sort_order_indicators]
```

spec/ransack/helpers/form_helper_spec.rb
jonatack: You're right, good catch, that test wasn't working :+1:

kalami:
I think I got everything.  Thanks for your patience!  This is my first contribution to a gem and I really appreciate the feedback and insights!

jonatack:
All good! 

Here is what remains: To keep the git history clean and clear, we need to merge the commits into one, but only the ones relevant to the new feature, and update the change log. The change that fixes the other spec, and the change to the `order_indicator` method in form_helper.rb, which have nothing to do with the new feature, should be in separate commits. For the moment we want one clean commit without the two other changes in it, and with an updated change log. 

We have two options. You can do it, but it might be a little complicated if you haven't squashed/cherry-picked/rebased commits before (I can help you if needed). Or I can do it. The result will be the same. Let me know!

jonatack:
P.S. It's great that working together, the bug in the other spec was found and fixed. Two pairs of eyes are better than one.

kalami:
If you don't mind, I'd rather leave that stuff up to you.  Thanks again!

jonatack:
I don't mind. Thanks for the contributions.

jonatack:
P.S. It looks like you have different user names in your github account and your `.gitconfig` file. The git history has commits from both Josh Hunter <josh-hunter@am.com> and from kalami <kalami@users.noreply.github.com>.

jonatack:
P.S. It looks like you have different user names/emails between your github account and your local `.gitconfig` file. The git history has commits from both `Josh Hunter <josh-hunter@am.com>` and from `kalami <kalami@users.noreply.github.com>`. Might be best to set them up the same.

jonatack:
Once your `.gitconfig` is aligned, would you like to make a quick PR to fix that other test you spotted?

PR #584, opened 2015-09-08
Escape underscore wildcard for pg and mysql.

IgorDobryn:
Escape mysql/pg `_` wildcard character.

### DIFF
--- a/lib/ransack/adapters/active_record/ransack/constants.rb
+++ b/lib/ransack/adapters/active_record/ransack/constants.rb
@@ -104,7 +104,7 @@ module Ransack
       case ActiveRecord::Base.connection.adapter_name
       when "Mysql2".freeze, "PostgreSQL".freeze
         # Necessary for PostgreSQL and MySQL
-        unescaped.to_s.gsub(/([\\|\%|.])/, '\\\\\\1')
+        unescaped.to_s.gsub(/([\\|\%|_|.])/, '\\\\\\1')
       else
         unescaped
       end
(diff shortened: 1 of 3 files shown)


a-chumagin:
+1

jonatack:
Thanks!

alkuzad:
+1 when this fix will be released ? It can be potentially dangerous

ernie:
FWIW, back when I built Ransack leaving wildcard matching supported was intentional. I think I even used to have a screencast demoing it. ;)

ernie:
But then, it was intended to be used for admin backends and the like. Folks seem to want to use it for more user-facing stuff than is wise.

alkuzad:
@ernie we faced that in LIKE statements (cont/end) when users can not match underscores, i.e `foo_end:"_bar"` will match anything that has bar at the end, not only the exact `_bar`

Great to know origins of this gem :) With the all that form helpers I always thought it's more user-oriented.

jonatack:
@ernie Interesting! Perhaps keep it as a config option, defaulting to 'safe'?

@alkuzad Yes, it's definitely time to cut a release. Til then, Ransack master works.

ernie:
@jonatack Totally up to you. To be fair, wildcards can return unexpected data but I wouldn't necessarily call them "unsafe". @alkuzad's concern, however, is warranted, in that if someone doesn't understand how to use wildcards, they may get confusing results.

alkuzad:
@ernie even if all our users would be devs they would not be possible to escape them as slash is also escaped :)

jonatack:
Thanks, this provides useful perspective. My own use has been for admin backends only, and more focused on simplicity for the users and fast live response with punctuation stripped out anyway.

Issue #589, opened 2015-09-17
Add support for alias_attribute


martndemus:
For example, I'm working on top of a legacy database which has all it's column names in german. I use `alias_attritbute` to map english names to the german column names. If I were to use Ransack I would have to support using the german attributes as query params.

With this example:

```ruby
class Location < ActiveRecord::Base
  self.table_name = 'orte'
  alias_attribute :code, :bezeichnung
end
```

The request `/api/locations?q[code_eq]=bremen` will not work. As it will construct the query as `where orte.code = 'bremen'`, while it should be translated to `where orte.bezeichnung = 'bremen'` .

jonatack:
Hello @martndemus, thank you for highlighting this. It seems like an edge case feature request rather than a bug with the current code. I do not plan to work on this feature request, but you are more than welcome to make a pull request to add it if it is worthwhile to you. Cheers!

PR #592, opened 2015-09-23
Add support for attribute alias

martndemus:
See #589

### DIFF
--- a/lib/ransack.rb
+++ b/lib/ransack.rb
@@ -8,6 +8,12 @@ Ransack::Adapters.require_constants
 module Ransack
   extend Configuration
   class UntraversableAssociationError < StandardError; end;
+
+  SUPPORTS_ATTRIBUTE_ALIAS = begin
+    ActiveRecord::Base.respond_to?(:attribute_aliases)
+  rescue NameError
+    false
+  end
 end
 
 Ransack.configure do |config|
--- a/lib/ransack/adapters/active_record/base.rb
+++ b/lib/ransack/adapters/active_record/base.rb
@@ -23,9 +23,12 @@ module Ransack
         # Ransackable_attributes, by default, returns all column names
         # and any defined ransackers as an array of strings.
         # For overriding with a whitelist array of strings.
-        #
         def ransackable_attributes(auth_object = nil)
-          column_names + _ransackers.keys
+          if Ransack::SUPPORTS_ATTRIBUTE_ALIAS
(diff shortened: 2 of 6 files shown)


lib/ransack/adapters/active_record/base.rb
```diff
@@ -23,9 +23,14 @@ def ransacker(name, opts = {}, &block)
...
         def ransackable_attributes(auth_object = nil)
-          column_names + _ransackers.keys
+          ransackable_attributes = column_names + _ransackers.keys
+
+          if Ransack::SUPPORTS_ATTRIBUTE_ALIAS
+            ransackable_attributes += attribute_aliases.keys
+          end
+
+          ransackable_attributes
         end
```
jonatack: We could probably just inline the values here without memoizing `column_names + _ransackers.keys` in a local var:

```ruby
if Ransack::SUPPORTS_ATTRIBUTE_ALIAS
  column_names + _ransackers.keys + attribute_aliases.keys
else
  column_names + _ransackers.keys
end
```

lib/ransack.rb
```diff
@@ -8,6 +8,12 @@
 module Ransack
   extend Configuration
   class UntraversableAssociationError < StandardError; end;
+
+  SUPPORTS_ATTRIBUTE_ALIAS = begin
+    ActiveRecord::Base.respond_to?(:attribute_aliases)
+  rescue NameError
+    false
+  end
 end
```
jonatack: Do you know if there are versions of Active Record that do not support `attribute_aliases`?

jonatack:
Thanks! This looks pretty good. Can you let me know for the question in the code?

lib/ransack.rb
martndemus: `attribute_aliases` was introduced in Rails 4. So 3.x versions do not support it. I haven't found an alternative method in Rails 3.x that retrieves the aliases.

jonatack: Perfect, thanks.

martndemus:
Inlined the things in the `ransackable_attributes` method.

jonatack:
Thanks! :green_apple:  Merged [here](https://github.com/activerecord-hackery/ransack/commit/549342aeed2a5e9962c43f14f744930511d86de8).

jonatack:
Would you like to make a PR to update the change log?

martndemus:
@jonatack:  see #594

Issue #483, opened 2014-12-11
Rails 4.2: NoMethodError: undefined method `asc' for #<Arel::Nodes::NamedFunction...


turboladen:
I'm in the process of upgrading a Rails 4.1.8 app to Rails 4.2.0.rc2.  I'm not seeing this behavior on 4.1.8 and I'm sure the error could be on my end, but I'm not quite sure I see a link to that.  Also, I've some wrappers (Concerns) around ransack that I'll try to unravel to simplify this report, so forgive me if things look a little funky.  Here goes...

Using MRI 2.1.0, I get the same results with master (ff0327df2c21) and the rails-4.2 branch; I've been using ransack 1.5.1 with my Rails 4.1.8 branch with success.

The backtrace for my failing specs all look similar to this:

```
Failure/Error: Unable to find matching line from backtrace
     NoMethodError:
       undefined method `asc' for #<Arel::Nodes::NamedFunction:0x007ffcf0ab4a88>
     Shared Example Group: "a pageable controller" called from ./spec/controllers/zone_maps_controller_spec.rb:32
     # /Users/sloveless/.gem/ruby/2.1.0/bundler/gems/ransack-ff0327df2c21/lib/ransack/visitor.rb:44:in `visit_Ransack_Nodes_Sort'
     # /Users/sloveless/.gem/ruby/2.1.0/bundler/gems/ransack-ff0327df2c21/lib/ransack/visitor.rb:52:in `visit'
     # /Users/sloveless/.gem/ruby/2.1.0/bundler/gems/ransack-ff0327df2c21/lib/ransack/visitor.rb:5:in `accept'
     # /Users/sloveless/.gem/ruby/2.1.0/bundler/gems/ransack-ff0327df2c21/lib/ransack/visitor.rb:13:in `block in visit_Array'
     # /Users/sloveless/.gem/ruby/2.1.0/bundler/gems/ransack-ff0327df2c21/lib/ransack/visitor.rb:13:in `map'
     # /Users/sloveless/.gem/ruby/2.1.0/bundler/gems/ransack-ff0327df2c21/lib/ransack/visitor.rb:13:in `visit_Array'
     # /Users/sloveless/.gem/ruby/2.1.0/bundler/gems/ransack-ff0327df2c21/lib/ransack/visitor.rb:52:in `visit'
[...]

PR #598, opened 2015-09-28
Fixed: Rails 4.2: NoMethodError: undefined method 'asc' for #<Arel::N…

roman-franko:
Fixed issue with Rails 4.2: NoMethodError: undefined method 'asc' for #<Arel::Nodes::NamedFunction... #483

Implemented wrapping of Arel::Nodes::NamedFunction for asc/desc case accordingly to Arel 6.0.

### DIFF
--- a/lib/ransack/visitor.rb
+++ b/lib/ransack/visitor.rb
@@ -44,7 +44,12 @@ module Ransack
     end
 
     def visit_Ransack_Nodes_Sort(object)
-      object.attr.send(object.dir) if object.valid?
+      return unless object.valid?
+      if object.attr.is_a? Arel::Attributes::Attribute
+        object.attr.send(object.dir)
+      else
+        ordered object
+      end
     end
 
     def quoted?(object)
@@ -64,5 +69,15 @@ module Ransack
       hash[klass] = "visit_#{klass.name.gsub('::', '_')}"
     end
 
+    private
+
+      def ordered(object)
+        case object.dir
+        when 'asc'.freeze
+          Arel::Nodes::Ascending.new(object.attr)
+        when 'desc'.freeze
+          Arel::Nodes::Descending.new(object.attr)
+        end
+      end
   end
 end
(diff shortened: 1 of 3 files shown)


jonatack:
@roman-franko Thanks! Could please you add a test that fails without this PR and passes with it?

lib/ransack/visitor.rb
```diff
@@ -44,7 +44,18 @@ def visit_or(object)
     end
 
     def visit_Ransack_Nodes_Sort(object)
-      object.attr.send(object.dir) if object.valid?
```
jonatack: This might be a good place for a guard conditional with early return like `return unless object.valid?`.

lib/ransack/visitor.rb
```diff
@@ -44,7 +44,18 @@ def visit_or(object)
...
+      if object.valid?
+        if object.attr.is_a? Arel::Attributes::Attribute
+          object.attr.send(object.dir)
+        else
+          case object.dir
```
jonatack: Perhaps extract this nested conditional into a separate method?

lib/ransack/visitor.rb
roman-franko: You are right, I've made some refactoring, thanks

lib/ransack/visitor.rb
```diff
@@ -64,5 +69,15 @@ def visit(object)
...
     end
 
+    private
+
+      def ordered(object)
+        case object.dir
+        when 'asc'
+          Arel::Nodes::Ascending.new(object.attr)
+        when 'desc'
+          Arel::Nodes::Descending.new(object.attr)
```
jonatack: LGTM. Could you freeze the two strings `'asc'` and `'desc'` please?

jonatack:
LGTM :+1:  If you could freeze the two strings (code comment above) and squash the commits that would be great. Otherwise I could do it if you don't have time, so just let me know. You might want to git fetch the latest version of master and rebase, it looks like you are using an out-of-date version of `base_spec.rb`.

roman-franko:
I've freezed strings for this branch. Actually rails-4.2 branch has some differences with master branch. I've duplicated all changes and made pull request to the master branch.

jonatack:
Thanks!

PR #599, opened 2015-09-29
Updating README with solution for DISTINCT issues

mikel:
By using activerecord `joins` with `includes`, or `select`, we can include the columns
the SQL engine needs to create the query, this solves a bunch of hackery.

* Added more tips on either using `joins` with `includes`, or using `select`
* Moved to it's own section, I missed this section entirely on first read through

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,43 +92,6 @@ If you're coming from MetaSearch, things to note:
   ActiveRecord::Relation in the case of the ActiveRecord adapter) via a call to
   `Ransack#result`.
 
-  4. If passed `distinct: true`, `result` will generate a `SELECT DISTINCT` to
-  avoid returning duplicate rows, even if conditions on a join would otherwise
-  result in some. It generates the same SQL as calling `uniq` on the relation.
-
-  Please note that for many databases, a sort on an associated table's columns
-  may result in invalid SQL with `distinct: true` -- in those cases, you're on
-  your own, and will need to modify the result as needed to allow these queries
-  to work.
-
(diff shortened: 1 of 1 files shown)


gurgelrenan:
This sounds really good!

jonatack:
Hi @mikel, thanks! I'm at a Ruby conference right now so sorry not to respond sooner. The README is becoming quite long to read. Since this is ultimately explaining how to use Active Record, I wonder if it might not be better to just mention the Ransack-specific `distinct: true` API and that it may have caveats, as before, with a link to various solutions all moved to a special section in the wiki.

mikel:
Sure, I think at a minimum it might be smart to put the .joins() and .includes() in there as the "simplest thing that could work" though, because lots of people trip up on it.

cllns:
Thanks, this just helped me! I think it's worth putting on the README, but at the very least it should be moved outside of this PR :)

Also, do you need to use `.joins` if you have `.includes`?

jonatack:
Ugh, it's late and I should be asleep instead of fixing merge conflicts. Merged this PR in, unfortunately messed up the PR credit; sorry about that.

I'm happy to be convinced otherwise, but my thoughts remains that the README has become bloated with passages that would best be in the wiki instead with links to it from the README -- particularly any parts about Active Record, Arel or SQL that are not specific to Ransack. So will likely simplify the README soon.

jonatack:
@mikel thanks again for improving the docs :heart:

PR #604, opened 2015-10-06
Add block argument for sort_link

dalpo:
Allow passing block argument to the sort_link method:

```ruby
<%= sort_link(@q, :name, [:name, 'kind ASC']) do %>
  <strong>Player Name</strong>
<% end %>
```

### DIFF
--- a/lib/ransack/helpers/form_helper.rb
+++ b/lib/ransack/helpers/form_helper.rb
@@ -38,12 +38,19 @@ module Ransack
       #
       #   <%= sort_link(@q, :name, [:name, 'kind ASC'], 'Player Name') %>
       #
-      def sort_link(search_object, attribute, *args)
+      # You can use a block as well if your link target is hard to fit into the label parameter:
+      #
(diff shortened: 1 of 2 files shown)


dalpo:
Uhm.. Failures don't depend on my commit :bug: 

It seems that the rack-cache gem has been changed and now require at least ruby 2.0

```
Gem::InstallError: rack-cache requires Ruby version >= 2.0.0.
Using activesupport 3.2.22 from git://github.com/rails/rails.git (at 3-2-stable)
An error occurred while installing rack-cache (1.3.0), and Bundler cannot
continue.
```

jonatack:
Thanks! Will have a look soon.

jonatack:
@dalpo Very sorry for not getting back to your PR much sooner. It looks great. Good job, mate. Would you like to make a PR to update the README and Change Log?

dalpo:
Sure! Here you can find the PR https://github.com/activerecord-hackery/ransack/pull/630

Issue #374, opened 2014-05-14
Searching across multiple associations that reference the same model/table joining on the wrong association


jhdavids8:
Hi all!

Let's say I have a model such as:

```
class Recommendation < ActiveRecord::Base
  belongs_to :user
  belongs_to :target_user, class_name: 'User'
end
```

Now, say my Ransack search filters looks like this (as JSON, as I'm using Ransack with a REST API. Hopefully, it's more readable):

```
q: {
  user: {
    job_title: { 
      eq: "Designer"
    }
  },
  target_user: {
    group: {
      name: {
        eq: "Design"
      }
    }
  }
}
```

When I was on Rails 4.0.2 and Ransack 1.2.1, this was the SQL output I received:

```
"SELECT \"recommendations\".* FROM \"recommendations\" 
LEFT OUTER JOIN \"users\" ON \"users\".\"id\" = \"recommendations\".\"user_id\" 
LEFT OUTER JOIN \"users\" \"target_users_recommendations\" ON \"target_users_recommendations\".\"id\" = \"recommendations\".\"target_user_id\" 
LEFT OUTER JOIN \"groups\" ON \"groups\".\"id\" = \"target_users_recommendations\".\"department_id\" 
WHERE ((\"users\".\"job_title\" = 'Designer' AND \"groups\".\"name\" = 'Design'))"
```

Notice how the group is joined here (4th line, last LEFT OUTER JOIN).

However, after upgrading to Rails 4.1.1 and the Ransack 4.1 branch, this is the SQL output, which is wrong, as notice now how the group is joined (it's joined on the first user association, not the target user association).

```
"SELECT \"recommendations\".* FROM \"recommendations\"
LEFT OUTER JOIN \"users\" ON \"users\".\"id\" = \"recommendations\".\"user_id\" 
LEFT OUTER JOIN \"users\" \"target_users_recommendations\" ON \"target_users_recommendations\".\"id\" = \"recommendations\".\"target_user_id\" 
LEFT OUTER JOIN \"groups\" ON \"groups\".\"id\" = \"users\".\"department_id\" 
WHERE ((\"users\".\"job_title\" = 'Designer' AND \"groups\".\"name\" = 'Design'))"
```

This is consistent across several specs I have failing at the moment. Any query on an association for the second user association will fail, as it joins on the former user association. 

Hopefully this is enough to get you guys started, but if not, please let me know. I'm going to try and find some time to dig deeper myself. Thanks!

[...]

PR #605, opened 2015-10-13
activerecord-hackery/ransack#374 Searching across multiple associatio…

bertBruynooghe:
…ns that reference the same model/table joining on the wrong association

### DIFF
--- a/spec/ransack/search_spec.rb
+++ b/spec/ransack/search_spec.rb
@@ -232,8 +232,7 @@ module Ransack
       # https://github.com/activerecord-hackery/ransack/issues/374
       #
       it 'evaluates conditions for multiple `belongs_to` associations to the
-      same table contextually',
-      if: ::ActiveRecord::VERSION::STRING.first(3) == '4.0' do
+      same table contextually' do
         s = Search.new(
           Recommendation,
           person_name_eq: 'Ernie',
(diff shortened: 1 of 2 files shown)


bertBruynooghe:
'Solved' the issue by dynamically creating associations on the models so the association code is solving the queries. Although it solves the issues I had and the tests that were introduced, I don't think it covers all code paths (polymorphic and or expressions are currently excluded), and it might render a few code paths into dead code...

jonatack:
@bertBruynooghe Thanks! Unfortunately travis-ci is red everywhere. Did the specs pass for you in your development environment?

bertBruynooghe:
The rails specs and ransack tests passed individually…

Regards,
Bert Bruynooghe.

jonatack:
OK, will give it a try in my local env.

spec/ransack/search_spec.rb
```diff
@@ -241,17 +240,23 @@ module Ransack
...
+           WHERE \\(+people.name = 'Ernie' AND (\\S+).name = 'Test'\\)+
+          SQL
+
+        match = Regexp.new(expected_query_exp.squish).match(real_query)
+
+        expect(match[1]).to eq match[2]
+        expect(match[1]).to eq match[5]
+        expect(match[3]).to eq match[4]
+        expect(match[3]).to eq match[6]
+        expect(match[1]).not_to eq match[3]
```
avit: We could do these assertions with Regexp backreferences.

Can we get a test here that executes the query and validates results to see it's valid?

avit:
There are only a few failing tests (but for every version it seems). It looks like something unrelated was broken as part of this change.

seanfcarroll:
I know this is old. If we want to merge it, it will need test coverage for the new methods.

Issue #39, opened 2011-11-02
Make alias to attributes


lest:
I am interested in making alias to set of attributes, e.g. instead of writing:

```
first_name_or_last_name_cont
```

it would be great to have ability to write simply:

```
full_name_cont
```

when somewhere we defines that full_name consists of first_name and last_name

PR #623, opened 2015-12-09
Allow creating aliases for ransack attributes

rzane:
Ransack is great, but the query strings can get crazy pretty quickly, especially when querying associations. It would be really great to have aliases for ransack attributes.

It seems like people want this feature built into Ransack. See:
+ #39
+ #253 
+ #536
+ https://github.com/jhdavids8/ransack-abbreviator

This PR allows you to do this:

```ruby
class Author < ActiveRecord::Base
  # Columns: (first_name, last_name)
end

  # Columns: (title, author_id)
  belongs_to :author
  ransack_alias :name, :title
  ransack_alias :author, :author_first_name_or_author_last_name
end

puts Post.ransack(name_start: 'Fun').result.to_sql
puts Post.ransack(author_cont: 'Ray').result.to_sql
```

```sql
SELECT "posts".* FROM "posts"
WHERE ("posts"."title" LIKE 'Fun%')

SELECT "posts".* FROM "posts"
LEFT OUTER JOIN "authors" ON "authors"."id" = "posts"."author_id"
WHERE ("authors"."first_name" LIKE '%Ray%' OR "authors"."last_name" LIKE '%Ray%')
```

### DIFF
--- a/lib/ransack/adapters/mongoid/base.rb
+++ b/lib/ransack/adapters/mongoid/base.rb
@@ -33,6 +33,14 @@ module Ransack
         end
 
         module ClassMethods
+          def _ransack_aliases
+            @_ransack_aliases ||= {}
+          end
+
+          def _ransack_aliases=(value)
+            @_ransack_aliases = value
+          end
+
           def _ransackers
             @_ransackers ||= {}
           end
@@ -49,13 +57,17 @@ module Ransack
 
           alias_method :search, :ransack
 
+          def ransack_alias(new_name, old_name)
+            self._ransack_aliases.store(new_name.to_s, old_name.to_s)
+          end
+
           def ransacker(name, opts = {}, &block)
             self._ransackers = _ransackers.merge name.to_s => Ransacker
               .new(self, name, opts, &block)
           end
 
           def all_ransackable_attributes
-            ['id'] + column_names.select { |c| c != '_id' } + _ransackers.keys
+            ['id'] + column_names.select { |c| c != '_id' } + _ransackers.keys + _ransack_aliases.keys
           end
 
           def ransackable_attributes(auth_object = nil)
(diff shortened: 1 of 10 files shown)


lib/ransack/adapters/mongoid/base.rb
```diff
@@ -33,6 +33,14 @@ def quote_column_name name
         end
 
         module ClassMethods
+          def _ransack_aliases
+            @_ransack_alises ||= {}
```
jonatack: Looks like a typo here. It did not cause the specs to fail?

jonatack:
Thanks for the PR! Will test it locally soon.

lib/ransack/adapters/mongoid/base.rb
rzane: Definitely a typo, but that variable is always being accessed through it's reader, which is why the tests pass anyway.

jonatack: Oh right, memoization. Thanks.

rzane:
Hey @jonatack, I fixed that typo. Have you had a chance to take a look?

jonatack:
LGTM, thanks @rzane. Could you update the README and Change Log to document this change?

rzane:
Sure. I'll do that right now.

jonatack:
Great.

lauramosher:
I have been excited for this change since I spotted it a month ago. Do you have an ETA for when this will be released?

jonatack:
Hi @lauramosher, you can use all the latest features now and at any time with the master branch.

We do need to do a release soon but there are a couple of issues to iron out before...

lauramosher:
@jonatack Thank you for the reply! Unfortunately, I am unable to use the master branch in this instance, but greatly look forward to the release. Keep up the awesome work.

laserlemon:
Looks great! Anything I can do to help move this into a release? Thank you! :clap:

PR #627, opened 2015-12-21
Add translations for equals, contains, starts_with, ends_with

istana:
See the issue https://github.com/activeadmin/activeadmin/issues/4150. I use activeadmin1.0.0pre2 and it generates long forms of predicates instead of short ones.


jonatack:
@istana Thank you for the PR. I believe this issue comes from ActiveAdmin adding custom predicate names to Ransack. Shouldn't it be resolved in ActiveAdmin?

istana:
@jonatack Thanks for the answer. I wasn't sure if it originates from ActiveAdmin or Ransack. I will ask in AA.

jonatack:
Sure. Let's close this for now, but ready to open it again if there's a good reason to handle it here. Thanks again.

PR #634, opened 2016-01-04
Rails 5 params: instance_variable_get -> to_unsafe_h

seanlinsley:
`search_params` didn’t need this behavior because it’s using the `@params` variable that’s already ready for use

### DIFF
--- a/lib/ransack/helpers/form_helper.rb
+++ b/lib/ransack/helpers/form_helper.rb
@@ -86,7 +86,7 @@ module Ransack
       class SortLink
         def initialize(search, attribute, args, params)
           @search         = search
-          @params         = parameters_hash(params)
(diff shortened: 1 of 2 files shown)


seanlinsley:
This updates the logic in ceafc0586a6929aa578ac0b1ad1f7c75219ca4c6 to use the public Rails API instead of accessing an internal instance variable

jonatack:
Thanks. `search_params` needed it when that commit was made, but Rails master has continued to change since then. Perhaps with [this PR](https://github.com/rails/rails/pull/22903) we'll be able to remove these patches.

prathamesh-sonpatki:
https://github.com/rails/rails/pull/22903 got merged :smile:

PR #644, opened 2016-02-02
Fix missing search params for sort link

ryanswood:
Problem:
- The controller params are no longer being passed into ransack's sort_link. The culprit is a recent change to `SortLinks` internal parameter hash. It is taking the controller's parameter object and converting it to a regular hash in which all the keys are strings. During the assembly of the sort link, the code is then attempting to extract the search params using the default key of `:q` (or whatever what was configured by the user). 

I added a test that is using the real AC::Parameter object so that this bug could be caught. I `fixed` my personal app by changing the search key to `'q'` though I feel that change is a hack until this PR has been addressed.

Bug introduced here: [989d](https://github.com/activerecord-hackery/ransack/commit/989d)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,6 +97,12 @@
 
     *Ethan Yang*
 
+*   Fix recent change to sort_links parameter_hash to account for symbol search
+    keys. Add test using real AC:Parameter object.
+    Commit [989d](https://github.com/activerecord-hackery/ransack/commit/989d)
+
+    *Ryan Wood*
+
 ### Changed
 
 *   Memory/speed perf improvement: Freeze strings in array global constants and
--- a/lib/ransack/helpers/form_helper.rb
+++ b/lib/ransack/helpers/form_helper.rb
@@ -122,7 +122,7 @@ module Ransack
 
           def parameters_hash(params)
             return params unless params.respond_to?(:to_unsafe_h)
-            params.to_unsafe_h
+            params.to_unsafe_h.with_indifferent_access
           end
 
           def extract_sort_fields_and_mutate_args!(args)
--- a/spec/ransack/helpers/form_helper_spec.rb
+++ b/spec/ransack/helpers/form_helper_spec.rb
@@ -311,10 +311,11 @@ module Ransack
       end
 
       context 'view has existing parameters' do
-        before do
-          @controller.view_context.params.merge!({ exist: 'existing' })
-        end
         describe '#sort_link should not remove existing params' do
+          before do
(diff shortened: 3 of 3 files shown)


jonatack:
Thanks! :purple_heart:  Will try to look at this soon. We'll need to keep the fix compatible with Rails 3 as well.

jonatack:
@ryanswood, with the current Rails master your test now seems to pass without the HWIA fix. Could you check if this is the case for you as well? If so, it may still be a good idea to merge in your AC::Parameters test.

jonatack:
@ryanswood Merged in a slightly modified version of your test spec and credited you in the change log. Thanks!

ryanswood:
@jonatack Sorry I have not been more reliable/available. I am at an early startup with just myself and another engineer. Regarding this PR, the fix was related to using Rails 4. I can't answer to Rails 5 as I am not using it yet (waiting for the final release). The reason being is that:

```ruby
return params unless params.respond_to?(:to_unsafe_h)
params.to_unsafe_h
```

in Rails 4, the second line would always be evaluated which would return a hash that has strings for keys, thus the attempt to pull off the `q` params using the default param key `:q` would fail to grab the available params. In test, a regular hash was being used which means it would not respond to

```ruby
ActionController::Parameters.new(q: 1).to_unsafe_h.fetch(:q)
=> KeyError: key not found: :q

ActionController::Parameters.new(q: 1).to_unsafe_h.fetch('q')
=> 1

ActionController::Parameters.new(q: 1).to_unsafe_h.with_indifferent_access.fetch(:q)
=> 1
```

Does the above make sense? It has been awhile since I had this all in my brain.

My memory is failing me now as to how the previous change to prepare for Rails 5 made this a problem for me. Let me know if I can be of assistance. I should be more available this time.

jonatack:
Thanks, you are correct. When the following is run...

```ruby
ActionController::Parameters.new(q: 1).to_unsafe_h.fetch(:q)
```

...with Rails 5 it works, but not with Rails 4.2.6. The reason is that in Rails 5, `#to_unsafe_h` calls the following method which converts hashes to HWIA, whereas in Rails 4 it just calls `#to_hash`.

```ruby
      def convert_parameters_to_hashes(value, using)
        case value
        when Array
          value.map { |v| convert_parameters_to_hashes(v, using) }
        when Hash
          value.transform_values do |v|
            convert_parameters_to_hashes(v, using)
          end.with_indifferent_access
        when Parameters
          value.send(using)
        else
          value
        end
      end
```

jonatack:
@ryanswood Hopefully this resolves it correctly. Thanks again.

Issue #381, opened 2014-06-02
has_many and negative predicates


Serpico84:
Is it possible somehow to use negative predicates like not_cont, not_in etc... on has_many associations?

Right now this doesn't seem to work in specific cases, for example in my application a User has_many Emails. What should i do if I want to find all users that do not have one or more "gmail.com" emails? (in other words, exclude from the results all users having at least a gmail.com address)

If i do a search like "q" => {:emails_address_not_cont => "gmail.com"}, this would exclude users that have ONLY gmail.com addresses, but will still include users with gmail.com emails if they have at least one email that's not gmail.com.

adisos:
I added this scope to my `User` model:
```ruby
def self.without_tag_id id
  join_sql = <<-SQL
    LEFT OUTER JOIN
    taggings taggings_exclude
    ON (
      taggings_exclude.taggable_id = users.id
      AND taggings_exclude.taggable_type = 'User'
    )
  SQL
  join_sql = join_sql.squish
  having_sql = <<-SQL
    NOT (
      ? = ANY(
        ARRAY_REMOVE(
          ARRAY_AGG(taggings_exclude.tag_id),
          NULL
        )
      )
    )
  SQL
  having_sql = having_sql.squish

  joins(join_sql).group('users.id').having(having_sql, id)
end
```
Then also added this method:
```ruby
def self.ransackable_scopes(auth_object = nil)
  %i(without_tag_id)
end
```

And then I can use this:
```ruby
User.ransack({ without_tag_id: 5})
```

avit:
I opened #645 to address this.

PR #645, opened 2016-02-16
Respect negative conditions for collection associations

avit:
When searching against attributes from a collection association (`has_many` / `:through` / `has_and_belongs_to_many`), a normal JOIN will include false positives when asking to exclude values (`!=`, `NOT IN`, `IS NOT NULL`) and another tuple is matched.

This is the basic case, see specs for details:

* A book exists with two tags: "medieval" and "fantasy"
* You search for books that are NOT "fantasy"
* The result would still include it, because it matches the other tag for NOT "fantasy"

In situations where this would happen, I've moved the `LEFT OUTER JOIN` into a subquery with inverted conditions. This is implemented with a correlated primary key which should help the query plan to avoid scanning entire tables in the subquery. The conditions now look like this:

```
WHERE root.id NOT IN (
  SELECT child.parent_id FROM child
   ... WHERE ... AND child.parent_id = root.id
)
```

Note that root.id is referenced in the inner query. The EXPLAIN query plan for this looks sensible, at least in MySQL where I explored it.

As far as I know this works with all standard ActiveRecord joins: I have not yet verified with composite keys or polymorphic joins or Nth-degree associations but it should be close to working if it isn't already.

I will start testing this against our own real-world data to see how it performs, please check it out too.

### Known issues:

- [x] Fix mongoid tests failures.
- [ ] Errors with `not_null` predicate: seems like a positive case and should use the original code branch.

### DIFF
--- a/lib/ransack/adapters/active_record.rb
+++ b/lib/ransack/adapters/active_record.rb
@@ -1,6 +1,8 @@
 require 'ransack/adapters/active_record/base'
 ActiveRecord::Base.extend Ransack::Adapters::ActiveRecord::Base
 
+require 'ransack/adapters/active_record/context'
+
 case ActiveRecord::VERSION::STRING
 when /^3\.0\./
   require 'ransack/adapters/active_record/3.0/context'
@@ -8,6 +10,4 @@ when /^3\.1\./
   require 'ransack/adapters/active_record/3.1/context'
 when /^3\.2\./
   require 'ransack/adapters/active_record/3.2/context'
-else
-  require 'ransack/adapters/active_record/context'
 end
--- a/lib/ransack/adapters/active_record/3.0/context.rb
+++ b/lib/ransack/adapters/active_record/3.0/context.rb
@@ -7,9 +7,11 @@ module Ransack
   module Adapters
     module ActiveRecord
       class Context < ::Ransack::Context
+
         # Because the AR::Associations namespace is insane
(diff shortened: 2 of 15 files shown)


avit:
Yes I'm fixing mongoid... :hammer:

jonatack:
Awesome :tada:  great work @avit :purple_heart:

jonatack:
Fixes #381.

avit:
@jonatack Thanks for the merge on this.

I have one more thing to add: the "NOT NULL" predicate still needs to be handled specially. I will add another PR for this.

janklimo:
Hey @avit thank you very much for your work on this. I was trying to use this functionality with conditional `has_many ... through` relationships with little success.

In my case:

- a Course `has_many :user_assigned_content_skills, -> { where(source: 'user') }, class_name: "ContentSkill"`
- and a ContentSkill `belongs_to :skill` and `belongs_to :course`

Then `Course.ransack({user_assigned_content_skills_skill_name_not_eq: 'ruby'}).result.to_sql` returns the following (i.e. false positives again if a course has multiple `content_skill`s):

`"SELECT courses.* FROM courses LEFT OUTER JOIN content_skills ON content_skills.course_id = courses.id AND content_skills.source = 'user' LEFT OUTER JOIN skills ON skills.id = content_skills.skill_id WHERE (skills.name != 'ruby')"`

I'm sorry to hijack this old thread but I was wondering if:

a) you know of a way to make this query work
b) there is a plan to have this implemented in Ransack?

Many thanks for any insights!

PR #665, opened 2016-04-27
Ensure conditions differing only by ransacker_args aren't filtered out by uniq!

ajporterfield:
We have a setup similar to http://ransack-demo.herokuapp.com/users/advanced_search.  We use ransackers + ransacker_args to create "virtual" attributes - by that I mean attributes that have the same name, but are passed different ransacker_args.  One scenario for us is total_donations_over_time, which is a ransacker that takes a date range as its ransacker_args.  We have "Total donations this year" and "Total donations last year" in our attributes dropdown, but they both map to total_donations_over_time.  Anyways, I found that subsequent conditions having the same attribute name, predicate, and value were getting removed by self.conditions.uniq! even if they had different ransacker_args.  This patch fixes the issue by passing a block to uniq!.

### DIFF
--- a/lib/ransack/nodes/grouping.rb
+++ b/lib/ransack/nodes/grouping.rb
@@ -44,7 +44,7 @@ module Ransack
             self.conditions << condition if condition.valid?
           end
         end
-        self.conditions.uniq!
+        remove_duplicate_conditions!
       end
       alias :c= :conditions=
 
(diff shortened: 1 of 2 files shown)


jonatack:
Thanks @ajporterfield for your PR! Sorry for the delay, will review it shortly.

jonatack:
Thanks again 🍰

PR #671, opened 2016-05-19
Update README.md

Fivell:
update examples with  valid syntax to  define class methods as private


jonatack:
Thank you! 💜 

(PS - Don't forget to use `[skip ci]` for documentation-only changes so that the large Travis CI matrix doesn't run needlessly.)

jonatack:
And apologies for not responding sooner 😬.

PR #682, opened 2016-06-02
Add list of predicates to readme

Paxa:
I often open this repo because I don't remember all predicates.
And for some predicates it was not obvious for me how to use it, so I had to search in repo.

I hope this section in README will be helpful for people like me, but probably need some help to make better description.

Also:
- `_eq_any` is same as `_in` ? (and `_not_eq_any` is same as `_not_in`?)
- What's the point to use `eq_all` ? If I use it as `col_eq: ["A", "B"]`, how something can be equal "A" and "B" at the same time?
- What's the point to use `gt_any`? If value in DB is greater then any then it's greater then smallest searchable value?
- Same with `gt_all` - if greater then all, then also greater then biggest
- If I do `start_all: ["A", "B"]`, anything can start with "A" and "B" at the same time?

### DIFF
--- a/README.md
+++ b/README.md
@@ -378,6 +378,35 @@ query parameters in your URLs.
 <% end %>
 ```
 
+### Search Matchers
+
+List of all possible predicates
+
+* `*_eq` - equal
+* `*_not_eq` - not equal
+* `*_match` - matches with `LIKE`, e.g. `q[email_matches]=%@gmail.com`
+* Also: `*_does_not_match`, `*_matches_any`, `*_matches_all`, `*_does_not_match_any`, `*_does_not_match_all`
+* `*_lt` - less then
+* `*_lteq` - less then or equal
+* `*_gt` - greater than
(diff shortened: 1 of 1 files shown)


jonatack:
Thanks @paxa, did you ever consult the [wiki](https://github.com/activerecord-hackery/ransack/wiki/Basic-Searching)?

jonatack:
Please see also the [CONTRIBUTING.md](https://github.com/activerecord-hackery/ransack/blob/master/CONTRIBUTING.md) about using `[skip ci]` for documentation-only commits. Thanks 😃

Paxa:
Thanks, I added link to wiki too

Paxa:
What do you think? Should it be merged?
Or just publish in wiki?

seanfcarroll:
There is already so much in the README that could also be in a wiki. Moving it into the wiki could be a future PR. Merging this.

PR #692, opened 2016-07-01
Fix returned value when several predicate.right

camelmasa:
I tried to change `spec/support/schema.rb` and `spec/ransack/adapters/active_record/base_spec.rb` like below.

```rb
# spec/support/schema.rb:69
ransacker :array_users,                                                                                                
  formatter: proc { |v| Person.where(id: v).map(&:id) } do |parent|                                                    
  parent.table[:id]                                                                                                    
end   
```

```rb
# spec/ransack/adapters/active_record/base_spec.rb:316
it 'should function correctly when passing an array of ids' do                                                           
  s = Person.ransack(array_users_in: [1, 2])                                                                               
  expect(s.result.count).to be 2
end
```

Should be pass tests but that couldn't pass tests.

Because, `Ransack::Nodes::Condition#format_predicate` method can target `predicate.right[0]` **only**.
I think the method should target several `predicate.right`.

### DIFF
--- a/lib/ransack/adapters/active_record/ransack/nodes/condition.rb
+++ b/lib/ransack/adapters/active_record/ransack/nodes/condition.rb
@@ -26,17 +26,23 @@ module Ransack
           arel_pred = arel_predicate_for_attribute(attribute)
           arel_values = formatted_values_for_attribute(attribute)
           predicate = attribute.attr.public_send(arel_pred, arel_values)
-          if casted_array_with_in_predicate?(predicate)
-            predicate.right[0] = format_values_for(predicate.right[0])
+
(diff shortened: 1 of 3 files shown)


camelmasa:
@jonatack Could you review when you have chance ❓  Thanks.

jonatack:
Thank you @camelmasa, tested it, LGTM.
:heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

jonatack:
Feel free to add a change log PR.

camelmasa:
@jonatack Thank you very much for merged this PR 🤗

Yeah, I'm going to add change log next week 👍

PR #672, opened 2016-05-22
Fix insecure params error and Add sort_url function.

amatotsuji:
I found error that `Generating a URL from non sanitized request parameters is insecure!`.
It caused by generating url using raw unpermited params.
Rails4 or 5 have strong parameters.
I fixed this problem.

And I add sort_url function.
It is return only url of sort_link.

Please review my code and merge to master.

PR #706, opened 2016-07-27
Add sort url

amatotsuji:
ref #672 
 I add sort_url function.
This is return only url of sort_link.
Also, I add test code for sort_url.


jonatack:
Thank you, @amatotsuji.

PR #717, opened 2016-08-26
Allow Active Record Ransack aliases with the same name on different models

laserlemon:
This fixes two bugs:

1. Subclasses were not properly inheriting their parents' Ransack
   aliases in Mongoid because each class defined its very own set of
   aliases.
2. In Active Record, Ransack aliases were defined in such a way that the
   parent's (and grandparent's, etc.) aliases were overwritten by the
   child, meaning that all aliases were ultimately kept on
   ActiveRecord::Base. This has the unfortunate effect of enforcing
   uniqueness of Ransack alias names across all models rather than per
   model. Depending on the load order of models, earlier definitions of
   an alias in other models would be clobbered.

### DIFF
--- a/lib/ransack/adapters/active_record/base.rb
+++ b/lib/ransack/adapters/active_record/base.rb
@@ -23,7 +23,8 @@ module Ransack
         end
 
         def ransack_alias(new_name, old_name)
-          self._ransack_aliases.store(new_name.to_s, old_name.to_s)
+          self._ransack_aliases = _ransack_aliases.merge new_name.to_s =>
+            old_name.to_s
         end
 
         # Ransackable_attributes, by default, returns all column names
--- a/lib/ransack/adapters/mongoid/base.rb
+++ b/lib/ransack/adapters/mongoid/base.rb
@@ -8,6 +8,10 @@ module Ransack
         extend ActiveSupport::Concern
 
         included do
+          class_attribute :_ransackers
+          class_attribute :_ransack_aliases
+          self._ransackers ||= {}
+          self._ransack_aliases ||= {}
         end
 
         class ColumnWrapper < SimpleDelegator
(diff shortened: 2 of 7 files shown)


laserlemon:
There's more background as to why this fix works in [Rails' documentation of `class_attribute`](https://github.com/rails/rails/blob/v5.0.0.1/activesupport/lib/active_support/core_ext/class/attribute.rb#L6-L70) (as of version 5.0.0.1).

lib/ransack/adapters/active_record/base.rb
```diff
@@ -23,7 +23,8 @@ def ransacker(name, opts = {}, &block)
         end
 
         def ransack_alias(new_name, old_name)
-          self._ransack_aliases.store(new_name.to_s, old_name.to_s)
+          self._ransack_aliases = _ransack_aliases.merge new_name.to_s =>
```
laserlemon: It's worth noting that this is exactly how the `ransacker` method already works.

lib/ransack/adapters/mongoid/base.rb
```diff
@@ -8,6 +8,10 @@ module Base
         extend ActiveSupport::Concern
 
         included do
+          class_attribute :_ransackers
+          class_attribute :_ransack_aliases
+          self._ransackers ||= {}
+          self._ransack_aliases ||= {}
```
laserlemon: Using `class_attribute` here makes much of Mongoid's implementation identical to that of Active Record's. :+1:

laserlemon:
💚 ❕

jonatack:
Thanks @laserlemon :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

Would you like to add a PR to update the change log? (Please use `[skip ci]` in the commit message if you do.)

PR #719, opened 2016-09-03
use `ActiveSupport.on_load` to hook into Active Record

y-yagi:
For avoiding autoloading these constants too soon.
This affect the initialization of Active Record.
Ref: https://github.com/rails/rails/issues/23589#issuecomment-229247727

### DIFF
--- a/lib/ransack.rb
+++ b/lib/ransack.rb
@@ -7,13 +7,6 @@ Ransack::Adapters.object_mapper.require_constants
 module Ransack
   extend Configuration
   class UntraversableAssociationError < StandardError; end;
-
-  SUPPORTS_ATTRIBUTE_ALIAS =
-  begin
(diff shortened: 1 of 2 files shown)


jonatack:
Thanks @y-yagi!

y-yagi:
@jonatack Thank you for your confirmation!

jdguzman:
I think this commit may have broken the mongoid adapter

jdguzman:
could we just add a bindable.rb to the mongoid adapter and override is_alias_attribute? to return false?

y-yagi:
Thank you pointed out! I think you are correct. I will fix.

jonatack:
Thank you @jdguzman for reporting.

ypresto:
This affects my rails 5 project, any plans for releasing this..?

PR #721, opened 2016-09-16
define `Ransack::SUPPORTS_ATTRIBUTE_ALIAS` to mongoid adapter

y-yagi:
In #719, the definition of `Ransack::SUPPORTS_ATTRIBUTE_ALIAS` has changed when
Active Record is loaded.

In the impact, `Ransack::SUPPORTS_ATTRIBUTE_ALIAS` in mongo adapter has become no longer be defined,
modify to defined in mongoid adapter.

Follow up to #719


jonatack:
Thank you, @y-yagi.

PR #726, opened 2016-09-27
Add customizable indicator arrows

garettarrowood:
This addition allows users to pass in their own up and down arrows. The default arrows remain the same. The user can pass in just one arrow, or both arrows.

Functionality is hand-tested with a codebase in production. Please let me know if you'd like any edits.

### DIFF
--- a/lib/ransack/helpers/form_helper.rb
+++ b/lib/ransack/helpers/form_helper.rb
@@ -110,11 +110,11 @@ module Ransack
         end
 
         def up_arrow
-          '&#9660;'.freeze
+          Ransack.options[:up_arrow]
         end
 
         def down_arrow
-          '&#9650;'.freeze
+          Ransack.options[:down_arrow]
         end
 
         def name
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Change Log
 
 ## Unreleased
+
+### Added
+
+*   Add config option to customize up and down arrows used for direction indicators in the FormHelper. PR [#726](https://github.com/activerecord-hackery/ransack/pull/726)
+
+    *Garett Arrowood*
+
 ### Fixed
 
 *   Use class attributes properly so that inheritance is respected.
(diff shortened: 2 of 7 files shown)


garettarrowood:
Sorry about the extra CI builds. Just realized it's been running a few times as I appended the documentation. Tried to log in and stop the last one to no avail. I'll be sure to include the [skip ci] option should a need to alter any documentation again.

lib/ransack/helpers/form_helper.rb
```diff
@@ -110,11 +110,11 @@ def initialize(search, attribute, args, params)
...
 
         def up_arrow
-          '&#9660;'.freeze
+          Ransack.options[:up_arrow]
         end
 
         def down_arrow
-          '&#9650;'.freeze
+          Ransack.options[:down_arrow]
         end
```
jonatack: I originally extracted these 2 public arrow methods out from the private method #order_indicator to allow developers to set custom arrows by overloading them.

Now that developers can do it with a config setting, I'm not sure there is any reason to keep them as separate public methods. Performance-wise it would be better to inline them.

WDYT?

jonatack: Oh, except if anyone is actually overloading these methods now, their apps will break. Never mind 😄

jonatack: Oh, except if anyone is actually overloading these methods now, their apps will break. Never mind 😄

jonatack:
@garettarrowood The changes were so minor that let's just merge this in and I'll do a mini-cleanup. Thanks again for the clean PR with tests.

garettarrowood:
@jonatack Cool! Thank you for the merge and cleaning up those comments/docs.

jonatack:
@garettarrowood Thanks again for your contribution!

garettarrowood:
Great refactoring and additions to the tests. I enjoyed reviewing your edits.

jonatack:
Thanks @garettarrowood.

I'm a bit bothered by all the tests duplicated for the two adapters. Code bloat. Probably should be fixed.

Issue #593, opened 2015-09-24
Parameter `1` automatically translated to `true`


adisos:
My workaround:
tag_id = 1 if tag_id == true

sjaveed:
Nice.  An alternate workaround (which I just put in place before looking here) is to pass in values as floats:
```ruby
class Post < ActiveRecord::Base
  scope :with_value, -> (val) { where(field_with_value: val.to_i) }

  def self.ransackable_scopes (auth_object = nil)
    %i/with_value/
  end
end
```
Use ransack like so:
```ruby
  Post.ransack({ with_value: 1.0 }).results
```

avit:
I think the plan should be to require Array for scope params without converting types, otherwise the value is just true/false for including the scope or not. See my comments on #509.

mostafa-zahran:
Another workaround is to make scope parameter as an optional parameter with default value 1 in case of ransack hash is dynamically created or stored in DB

seanfcarroll:
Closed by https://github.com/activerecord-hackery/ransack/pull/742

Issue #580, opened 2015-09-02
Searches with scopes fail when the scope has a single argument and integer 1 is passed as a parameter


AndrewSwerlick:
Here's a fun little bug that had me scratching my head for a while. If you try to do a search using a scope which takes a single parameter, your search will fail if you try to use the value 1 as the parameter. Failing test case here https://github.com/AndrewSwerlick/ransack/blob/scope_bug/spec/ransack/adapters/active_record/base_spec.rb#L66

The test errors out with 

```
 ArgumentError:
      wrong number of arguments (0 for 1)
    # ./spec/support/schema.rb:48:in `block in <class:Person>'
    # ./lib/ransack/context.rb:50:in `public_send'
    # ./lib/ransack/context.rb:50:in `chain_scope'
    # ./lib/ransack/search.rb:130:in `add_scope'
    # ./lib/ransack/search.rb:45:in `block in build'
    # ./lib/ransack/search.rb:39:in `each'
    # ./lib/ransack/search.rb:39:in `build'
    # ./lib/ransack/search.rb:31:in `initialize'
    # ./lib/ransack/adapters/active_record/base.rb:15:in `new'
    # ./lib/ransack/adapters/active_record/base.rb:15:in `ransack'
    # ./spec/ransack/adapters/active_record/base_spec.rb:67:in `block (4 levels) in <module:ActiveRecord>'
```

I believe the problem actually stems from https://github.com/activerecord-hackery/ransack/blob/master/lib/ransack/search.rb#L138. Basically the sanitized_scope_args method converts the 1 to a true, which then changes how it's handled by `Context#chain_scope`.

This issue appears to have been introduced in 1.6.0, as it only cropped up in my application since we upgraded from 1.5.1. The santizied_scope_args method does not appear to existing in 1.5.1.

kwerle:
This has bit me, too.  Reverting to 1.5.1 does the trick.

adisos:
https://github.com/activerecord-hackery/ransack/issues/593

haobole:
:+1:

bolshakov:
:+1:

bolshakov:
I think this should be fixed in ransack. My current fix:

```ruby
constants_to_remove = Set.new([1, '1', 't', 'T'])
Ransack::Constants::TRUE_VALUES = Ransack::Constants::TRUE_VALUES - constants_to_remove
[...]

PR #575, opened 2015-08-18
Passing boolean values to scopes

MarcelEeken:
Currently it is not possible to pass boolean values to scopes, see issue #566 .
This are two tests cases for those situations.

Issue #566, opened 2015-07-15
Getting unexpected coercion within scopes.


alejandrodevs:
I know that it is a discussed issue but I'm still getting errors.
Example:

``` ruby
class Payment < ActiveRecord::Base
  scope :by_month, -> (month) { where('extract(month from created_at) = ?', month) }
  def self.ransackable_scopes(object)
    [:by_month]
  end
end

Payment.ransack(by_month: 1)     # I get true param in scope.
Payment.ransack(by_month: '1')    # I get true param in scope.
Payment.ransack(by_month: [1])   # I get true param in scope.
Payment.ransack(by_month: ['1'])  # I get true param in scope.
```

ruby 2.2.2
rails 4.2.1
ransack 1.6.6

I want to catch 1 or '1' as param. Any idea?

MarcelEeken:
I have the same problem currently here.

Where i expect value such as 1 or true to be passed as string to my scope, they are converted to a boolean true. Even if they are enclosed in an array.

MarcelEeken:
While looking further in to this I discovered that the way you check for scope arity doesn't work in rails 4.
You check scope arity by using the following method: https://github.com/activerecord-hackery/ransack/blob/master/lib/ransack/context.rb#L53 which uses the arity method of a ruby method object.

If you look how scopes are implemented currently: https://github.com/rails/rails/blob/b1879124a82b34168412ac699cf6f654e005c4d6/activerecord/lib/active_record/scoping/named.rb#L154 you see that is sends a method, where the block passed is the scope method. So when you ask arity for the scope, you are actually asking for arity of the scope block from rails, which is always -1.

jonatack:
@MarcelEeken thanks for looking into this. I use ransackers as they were the original implementation by Ernie for doing this. Scopes in Ransack were added much later by the community but never quite finished/polished. The early users of Ransack probably mostly stayed with their ransackers, but I wouldn't be surprised if many newer users are using scopes. So, tests and PRs for using scopes are very welcome :smiley:

jonatack:
I believe that yes, this issue has been discussed in depth at least once before. Do have a look, and don't hesitate to link this issue to the previous discussion.

MarcelEeken:
I started looking into it, but haven't found a solution yet. I have a red test case, but have not been able to think of a solution that doesn't clash with the way of conditionally adding a scope with a true or false value.

jonatack:
[...]

PR #742, opened 2016-12-11
Config option for turning off custom scope argument sanitization

garettarrowood:
There has been a lot of discussion about ways to get around Ransack converting scope values to booleans. Here are just a few places: #509, #593, #580, #566.  This PR specifically addresses the tests presented in #575.

An easy solution to this would be to simply stop making these conversions. But then we run the risk of breaking backwards compatibility. So instead of changing them, I've added a config option to turn them off.

It'd be easy to modify this to make the config option more granular. We could set it up for the user to specify what exact truthy/falsey values they would like converted, and use the default settings if this config is NOT specified. For example, instead of:

```
    # Ransack.configure do |config|
    #   # Accept my custom scope values as what they are.
    #   config.sanitize_custom_scope_booleans = false
    # end
```

we could do this:

```
    # Ransack.configure do |config|
    #   # Only convert these values to booleans
    #   config.truthy_values_to_convert_in_custom_scopes = ['TRUE', 'true', '1']
    #   config.falsey_values_to_convert_in_custom_scopes = ['FALSE', 'no way no how']
    # end
```

But I don't know if that level of detail is needed/desired. If being able to disable boolean conversion is not enough, this could be added later on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@
 
     *Garett Arrowood*
 
+*   Add ability to turn off sanitization of custom scope arguments.
+    PR [#742](https://github.com/activerecord-hackery/ransack/pull/742).
+
+    *Garett Arrowood*
+
 ### Fixed
 
 *   Use class attributes properly so that inheritance is respected.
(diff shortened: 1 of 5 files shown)


truonglocbinh93:
@garettarrowood I added your config but error when start server Is there any solution for this problem? Please help me.

stevebissett:
This was really helpful today. Thank you @garettarrowood 🏆

PR #755, opened 2017-01-27
raise exception for unsupported adapters

timoschilling:
The reason can be found here: https://github.com/activeadmin/activeadmin/issues/4756
The solution is from @deivid-rodriguez in: https://github.com/activeadmin/activeadmin/issues/4756#issuecomment-275505455


timoschilling:
The tests failing because of a dependency problem between ruby 2.0 and nokogiri.

seanfcarroll:
Thanks @timoschilling

PR #762, opened 2017-02-09
[#138938207] Add scope support for ransortable_attributes

drborges:
Allows sorting logic to be applied through `ActiveRecord` scopes rather than just column names. Ransack already supports scoped filtering through `ransackable_scopes`, this PR provides a similar behavior for `ransortable_attributes` (see Ransack [README](https://github.com/activerecord-hackery/ransack#authorization-whitelistingblacklisting) for more information on these features).

### More Background and Use-cases

This change will allow for sorting on dynamically computed columns defined via scopes.

For example, given the `LoanApplication` model below defining some sorting scopes:

```ruby
class LoanApplication < ActiveRecord::Base
  # Sorts loan applications by the approval rate computed below
  scope :sort_by_approval_rate, ->(dir: "ASC") { order("(approved_amount / applied_amount) #{dir}") }
  scope :sort_by_approval_rate_asc, -> { sort_by_approval_rate }
  scope :sort_by_approval_rate_desc, -> { sort_by_approval_rate dir: "DESC" }

  class << self
    def ransortable_attributes(auth_object = nil)
      super(auth_object) + [
        :sort_by_approval_rate_asc,
        :sort_by_approval_rate_desc,
      ]
    end
  end
end
```

When using `Ransack` to search for loan applications with the criteria of `params[:q] = { s: "approval_rate asc" }`

```ruby
class LoanApplicationController  < ApplicationController
  def index
    @loan_applications = LoanApplication.ransack(params[:q]).result
  end
end
```

Ransack will verify that `approval_rate` is not a valid column on the `loan_applications` table, and will then check if the model defines sorting scopes for `approval_rate` and use them instead.

### Naming sorting scopes:

Sorting scopes are named as `sort_by_#{scope_name}_#{sorting_direction}`, where `scope_name` in the example above is `approval_rate` and `sorting_direction` is `asc` (define as the `s` attribute in the query `q` from `params`.

### Credits

This solution was borrowed from [this gist](https://gist.github.com/laserlemon/e5ce4e0d8bac63f4aa6835605ce3ec21).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Change Log
 
 ## Unreleased
+
+### Added
+
+*   Add config option to customize up and down arrows used for direction indicators in the FormHelper. PR [#726](https://github.com/activerecord-hackery/ransack/pull/726)
+
+    *Garett Arrowood*
+
 ### Fixed
 
 *   Use class attributes properly so that inheritance is respected.
--- a/README.md
+++ b/README.md
@@ -213,8 +213,18 @@ The sort link may be displayed without the order indicator arrow by passing
 <%= sort_link(@q, :name, hide_indicator: true) %>
 ```
 
-Alternatively, all sort links may be displayed without the order indicator arrow
-by adding this to an initializer file like `config/initializers/ransack.rb`:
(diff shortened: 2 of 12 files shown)


drborges:
My apologies, this PR was not meant to be opened upstream just yet.

PR #766, opened 2017-02-14
Consider ransack aliases when sorting

faragorn:
`ransack_alias` is a great feature however it does not work with sorting out of the box. In my opinion `ransack_aliases` should be recognized by sort function out of the box. 

Consider the following example:
```ruby
  belongs_to :author

  # Abbreviate :author_first_name_or_author_last_name to :author
  ransack_alias :author, :author_last_name
end
```
 In your view:
```
  <%= sort_link(@q, :author, 'Author Name') %>
```

### DIFF
--- a/lib/ransack/nodes/sort.rb
+++ b/lib/ransack/nodes/sort.rb
@@ -30,9 +30,9 @@ module Ransack
         .include?(attr_name)
       end
 
-      def name=(name)
-        @name = name
-        context.bind(self, name)
+      def name=name
+        @name = context.ransackable_alias(name) || name
+        context.bind(self, @name)
       end
 
       def dir=(dir)


seanfcarroll:
Hi @faragorn this looks good and should merge now we not supporting Ruby 2. Could you write some tests to cover this before it is merged?

ekampp:
Hey, @seanfcarroll, I need this feature, so if you could help me understand, where I should start attempting to add test coverage, I will open a new PR with this code and test coverage. Thanks.

deivid-rodriguez:
Closing as this was fixed by #1223.

@faragorn Unfortunately you were not credited as commit author even if you are the original author of this solution, I will add your name to the changelog though! Thanks!

PR #791, opened 2017-05-17
Fix markdown formatting in README

MattFenelon:
* The badges at the top of the README are showing the links as text, rather than being hyperlinks.
* Some of the headers don't have spaces between the #s and the text, causing the #s to be rendered.


MattFenelon:
Oops. I didn't notice the other PRs before I created this one.

PR #811, opened 2017-06-30
Add support to baby_squeel gem

thiagopnobre:
Fixes Ransack::Adapters::ActiveRecord::Context class, to add support to baby_squeel gem.

### DIFF
--- a/lib/ransack/adapters/active_record/context.rb
+++ b/lib/ransack/adapters/active_record/context.rb
@@ -260,6 +260,8 @@ module Ransack
               :stashed_join
             when Arel::Nodes::Join
               :join_node
+            when BabySqueel::JoinExpression
+              :joining
             else
               raise 'unknown class: %s' % join.class.name
             end


seanfcarroll:
I've got mixed thoughts on this: 

- do we want to further complicate the Ransack API? I'm not a big fan of having supported Mongo
- No matter what, we would need a thorough test suite for BabySqueel specific queries
- There is still a lot of cleanup / documentation to be done on the existing functionality 

What is the consensus from the community?

ernie:
The adapters were always intended to allow support of multiple versions of active record in a more maintainable way. This is a tiny change, but feels like a slippery slope, and something that doesn't really belong in the gem, and should be monkey patched in. Y folks who want to use it with baby squeel. Additionally, I'm kind of surprised the Travis build passed. We probably need a test to exercise the unknown node case, since as is, this would throw a confusing error on the missing constant for those who don't have baby squeel in their app.

thiagopnobre:
Yeah, I understand your concerns about how much more work would be needed to realy assure that the ransack is really fully compatible with the baby_squeel gem. But on my first thoughts, I considered the possibility because the baby_squeel is a gem with a sintax squeel like, but with less monkeypatches, and also because this tiny change worked well within a project that I'm working on (I'm still considering and testing the use of the baby_squeel gem in it, since squeel it's apparently not being manteined anymore).

So I opened this PR on the intention to make it possible to work with these two gem simutaneously, since I can't stop to use ransack because of this imcompatibility, forcing me to make a fork to test baby_squeel that is a cool gem, but has this integration problem with ransack.

I know that you have to make a lot of considerations to see if it's really a good idea to go through with this or not to assure that this two gems are really fully compatible. So my intention here it's to at least raising up some questioning about it.

Thanks for the feedback and sorry for the english :)

seanfcarroll:
@thiagopnobre  thank you for your contribution. Can we close this PR then?

thiagopnobre:
Yes, I will close it.

PR #818, opened 2017-07-28
fix bug in sort_link helper around argument extraction about sort fields with block.

smoriwaki:
fix bug in sort_link helper around argument extraction about sort fields with block.

### DIFF
--- a/lib/ransack/helpers/form_helper.rb
+++ b/lib/ransack/helpers/form_helper.rb
@@ -49,7 +49,7 @@ module Ransack
         unless Search === search
           raise TypeError, 'First argument must be a Ransack::Search!'
         end
-        args.unshift(capture(&block)) if block_given?
+        args[args.first.is_a?(Array) ? 1 : 0, 0] = capture(&block) if block_given?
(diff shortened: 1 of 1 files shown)


seanfcarroll:
Hi @smoriwaki thank you for the PR.

Is it possible to provide more detail around the underlying issue, as well as a failing and then passing test? Thanks!

smoriwaki:
This pull request fix the bug in sort_link, which causes the multiple fields option to be ignored when block parameter is specified. For example in the following case,

```
<%= sort_link(@q, :last_name, %i(last_name first_name), default_order: {last_name: 'asc', first_name: 'desc'}) do %>
  Full Name
<% end %>
```

the following result is expected,

```
<a class="sort_link " href="/some_resources?q%5Bs%5D%5B%5D=first_name+asc&amp;q%5Bs%5D%5B%5D=last_name+asc">Full Name</a>
```

but, actually the following result is got.

```
<a class="sort_link " href="/some_resources?q%5Bs%5D=last_name+asc">Full Name</a>
```

Issue #819, opened 2017-07-29
Improving memory efficiency


seanlinsley:
How can we optimize Ransack's memory usage? I'd like to discuss that here, ideally with multiple PRs created as a result of that discussion.

I discovered with the rack-mini-profiler and memory_profiler gems that Ransack creates a lot of objects in memory while rendering a page, just to be thrown away. I'd started looking into ActiveAdmin memory usage, but discovered that Ransack creates twice the number of objects as ActiveAdmin does.

https://github.com/activeadmin/activeadmin/issues/5113

You can download the [full report](https://github.com/activerecord-hackery/ransack/files/1185151/ransack-memory.txt), but here are the highlights:
```
Total allocated: 1401988 bytes (17704 objects)
Total retained:  1112 bytes (18 objects)

allocated memory by location
-----------------------------------
    462768  ransack-8ca8c95c6543/lib/ransack/predicate.rb:28
    215840  ransack-8ca8c95c6543/lib/ransack/adapters/active_record/base.rb:36
    129708  ransack-8ca8c95c6543/lib/ransack/predicate.rb:22
    120080  ransack-8ca8c95c6543/lib/ransack/predicate.rb:13
    113760  ransack-8ca8c95c6543/lib/ransack/predicate.rb:9
    107920  ransack-8ca8c95c6543/lib/ransack/adapters/active_record/base.rb:37
     98448  ransack-8ca8c95c6543/lib/ransack/adapters/active_record/base.rb:48
     30552  ransack-8ca8c95c6543/lib/ransack/adapters/active_record/ransack/context.rb:41


allocated objects by location
-----------------------------------
     11190  ransack-8ca8c95c6543/lib/ransack/predicate.rb:28
      2058  ransack-8ca8c95c6543/lib/ransack/adapters/active_record/base.rb:48
       730  ransack-8ca8c95c6543/lib/ransack/predicate.rb:22
       568  ransack-8ca8c95c6543/lib/ransack/adapters/active_record/base.rb:36
       330  ransack-8ca8c95c6543/lib/ransack/context.rb:138
       316  ransack-8ca8c95c6543/lib/ransack/predicate.rb:13
       284  ransack-8ca8c95c6543/lib/ransack/adapters/active_record/base.rb:37
```

This is the query string I used on an ActiveAdmin index page of my app:
```
?pp=profile-memory&memory_profiler_allow_files=lib/ransack
```
[...]

PR #820, opened 2017-08-02
Improve memory usage

seanlinsley:
Proof-of-concept for #819

These changes decreased memory usage by 96% 🚀 (1.3 MB to 58 KB)

```
Total allocated: 59848 bytes (1164 objects)
Total retained:  712 bytes (14 objects)

allocated memory by location
-----------------------------------
      8048  ransack/lib/ransack/adapters/active_record/ransack/context.rb:11
      7056  ransack/lib/ransack/adapters/active_record/context.rb:55
      4080  ransack/lib/ransack/adapters/active_record/context.rb:49
      3024  ransack/lib/ransack/adapters/active_record/context.rb:50
      2872  ransack/lib/ransack/adapters/active_record/context.rb:254
      2680  ransack/lib/ransack/adapters/active_record/context.rb:268
      2680  ransack/lib/ransack/adapters/active_record/context.rb:271

allocated objects by location
-----------------------------------
       102  ransack/lib/ransack/adapters/active_record/context.rb:49
        68  ransack/lib/ransack/adapters/active_record/context.rb:254
        67  ransack/lib/ransack/adapters/active_record/context.rb:268
        67  ransack/lib/ransack/adapters/active_record/context.rb:271
        67  ransack/lib/ransack/adapters/active_record/context.rb:272
        67  ransack/lib/ransack/adapters/active_record/context.rb:278
        67  ransack/lib/ransack/adapters/active_record/context.rb:284
```

### DIFF
--- a/lib/ransack/adapters/active_record/base.rb
+++ b/lib/ransack/adapters/active_record/base.rb
@@ -32,7 +32,7 @@ module Ransack
         # For overriding with a whitelist array of strings.
         #
         def ransackable_attributes(auth_object = nil)
-          if Ransack::SUPPORTS_ATTRIBUTE_ALIAS
+          @ransackable_attributes ||= if Ransack::SUPPORTS_ATTRIBUTE_ALIAS
             column_names + _ransackers.keys + _ransack_aliases.keys +
             attribute_aliases.keys
           else
@@ -45,7 +45,7 @@ module Ransack
         # For overriding with a whitelist array of strings.
         #
         def ransackable_associations(auth_object = nil)
-          reflect_on_all_associations.map { |a| a.name.to_s }
+          @ransackable_associations ||= reflect_on_all_associations.map { |a| a.name.to_s }
         end
 
         # Ransortable_attributes, by default, returns the names
--- a/lib/ransack/adapters/active_record/ransack/context.rb
+++ b/lib/ransack/adapters/active_record/ransack/context.rb
@@ -37,15 +37,14 @@ module Ransack
         @base = @join_dependency.join_base
         @engine = @base.arel_engine
       end
+    end
 
-      @default_table = Arel::Table.new(
-        @base.table_name, as: @base.aliased_table_name, type_caster: self
-        )
-      @bind_pairs = Hash.new do |hash, key|
-        parent, attr_name = get_parent_and_attribute_name(key)
-        if parent && attr_name
-          hash[key] = [parent, attr_name]
-        end
+    def bind_pair_for(key)
+      @bind_pairs ||= {}
+
+      @bind_pairs[key] ||= begin
+        parent, attr_name = get_parent_and_attribute_name(key.to_s)
+        [parent, attr_name] if parent && attr_name
       end
     end
 
--- a/lib/ransack/configuration.rb
+++ b/lib/ransack/configuration.rb
@@ -5,7 +5,27 @@ module Ransack
   module Configuration
 
     mattr_accessor :predicates, :options
-    self.predicates = {}
+
+    class PredicateCollection
+      attr_reader :sorted_names_with_underscores
+
+      def initialize
+        @collection = {}
+        @sorted_names_with_underscores = []
+      end
+
+      delegate :[], :keys, :has_key?, to: :@collection
+
+      def []=(key, value)
+        @sorted_names_with_underscores << [key, '_' + key]
+        @sorted_names_with_underscores.sort! { |(a, _), (b, _)| b.length <=> a.length }
+
+        @collection[key] = value
+      end
+    end
+
+    self.predicates = PredicateCollection.new
+
     self.options = {
       :search_key => :q,
       :ignore_unknown_conditions => true,
--- a/lib/ransack/context.rb
+++ b/lib/ransack/context.rb
@@ -40,7 +40,7 @@ module Ransack
     # Convert a string representing a chain of associations and an attribute
     # into the attribute itself
     def contextualize(str)
-      parent, attr_name = @bind_pairs[str]
+      parent, attr_name = bind_pair_for(str)
       table_for(parent)[attr_name]
     end
 
@@ -59,24 +59,24 @@ module Ransack
 
     def bind(object, str)
       return nil unless str
-      object.parent, object.attr_name = @bind_pairs[str]
+      object.parent, object.attr_name = bind_pair_for(str)
     end
 
     def traverse(str, base = @base)
       str ||= ''.freeze
 
-      if (segments = str.split(/_/)).size > 0
+      if (segments = str.split(Constants::UNDERSCORE)).size > 0
         remainder = []
         found_assoc = nil
         while !found_assoc && segments.size > 0 do
           # Strip the _of_Model_type text from the association name, but hold
           # onto it in klass, for use as the next base
           assoc, klass = unpolymorphize_association(
-            segments.join('_'.freeze)
+            segments.join(Constants::UNDERSCORE)
             )
           if found_assoc = get_association(assoc, base)
             base = traverse(
-              remainder.join('_'.freeze), klass || found_assoc.klass
+              remainder.join(Constants::UNDERSCORE), klass || found_assoc.klass
               )
           end
 
@@ -93,9 +93,9 @@ module Ransack
       base = klassify(base)
       str ||= ''.freeze
       path = []
-      segments = str.split(/_/)
+      segments = str.split(Constants::UNDERSCORE)
       association_parts = []
-      if (segments = str.split(/_/)).size > 0
+      if (segments = str.split(Constants::UNDERSCORE)).size > 0
         while segments.size > 0 &&
         !base.columns_hash[segments.join(Constants::UNDERSCORE)] &&
         association_parts << segments.shift do
@@ -135,7 +135,7 @@ module Ransack
     end
 
     def ransackable_scope?(str, klass)
-      klass.ransackable_scopes(auth_object).any? { |s| s.to_s == str }
+      klass.ransackable_scopes(auth_object).any? { |s| s.to_sym == str.to_sym }
     end
 
     def searchable_attributes(str = ''.freeze)
--- a/lib/ransack/predicate.rb
+++ b/lib/ransack/predicate.rb
@@ -9,34 +9,26 @@ module Ransack
         Ransack.predicates.keys
       end
 
-      def names_by_decreasing_length
-        names.sort { |a, b| b.length <=> a.length }
-      end
-
       def named(name)
         Ransack.predicates[name.to_s]
       end
 
       def detect_and_strip_from_string!(str)
-        if p = detect_from_string(str)
-          str.sub! /_#{p}$/, ''.freeze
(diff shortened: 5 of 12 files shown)


lib/ransack/adapters/active_record/base.rb
```diff
@@ -45,7 +45,7 @@ def ransackable_attributes(auth_object = nil)
         # For overriding with a whitelist array of strings.
         #
         def ransackable_associations(auth_object = nil)
-          reflect_on_all_associations.map { |a| a.name.to_s }
+          @ransackable_associations ||= reflect_on_all_associations.map { |a| a.name.to_s }
```
ernie: You'll want to be careful about memoization in these -- that's why they weren't to begin with. Technically, new values could be added by code at runtime -- they would need to invalidate the cache.

seanlinsley: Yeah, I hesitated to make this change. This PR isn't meant to be mergeable in its current state, I just went through the code and applied optimizations to the code that the report highlighted.

Now that we have this proof-of-concept, we can discuss which changes make sense.

avit: Possibly:

```
if @ransackable_associations && @ransackable_associations.size == reflect_on_all_associations.size
  @ransackable_associations
else
  @ransackable_associations = reflect_on_all_associations.map { |a| a.name.to_s }
end
```

jaredbeck: > Technically, new values could be added by code at runtime -- they would need to invalidate the cache.

Depending on your definition of "runtime", `PaperTrail` does this (adds associations to your models).

lib/ransack/adapters/active_record/ransack/context.rb
```diff
@@ -37,15 +37,14 @@ def initialize(object, options = {})
         @base = @join_dependency.join_base
         @engine = @base.arel_engine
       end
+    end
 
-      @default_table = Arel::Table.new(
-        @base.table_name, as: @base.aliased_table_name, type_caster: self
-        )
```
seanlinsley: This variable didn't seem to be used anywhere and it represented a lot of memory usage, so I removed it.

lib/ransack/adapters/active_record/ransack/context.rb
```diff
@@ -37,15 +37,14 @@ def initialize(object, options = {})
...
-      @bind_pairs = Hash.new do |hash, key|
-        parent, attr_name = get_parent_and_attribute_name(key)
-        if parent && attr_name
-          hash[key] = [parent, attr_name]
-        end
+    def bind_pair_for(key)
+      @bind_pairs ||= {}
+
+      @bind_pairs[key] ||= begin
+        parent, attr_name = get_parent_and_attribute_name(key.to_s)
```
seanlinsley: It's not clear to me why this would help (other than not initializing a hash until a bind pair is actually needed), but it represented 18 KB, while the line above it represented 30 KB.

lib/ransack/configuration.rb
```diff
@@ -5,7 +5,27 @@ module Ransack
...
+
+      def []=(key, value)
+        @sorted_names_with_underscores << [key, '_' + key]
+        @sorted_names_with_underscores.sort! { |(a, _), (b, _)| b.length <=> a.length }
+
+        @collection[key] = value
+      end
+    end
+
+    self.predicates = PredicateCollection.new
```
seanlinsley: This optimization makes it so that we don't have to re-sort the predicate array numerous times, and so that we don't have to re-allocate strings for checking `end_with?` in `predicate.rb`.

This change has the biggest impact, saving 806 KB (4 of the top 5 offenders)

seanlinsley: The problem is, this code assumes that no object will modify the predicates hash in-place.

seanlinsley: It also assumes that the same predicate key is never set twice.

lib/ransack/context.rb
```diff
@@ -135,7 +135,7 @@ def ransackable_association?(str, klass)
     end
 
     def ransackable_scope?(str, klass)
-      klass.ransackable_scopes(auth_object).any? { |s| s.to_s == str }
+      klass.ransackable_scopes(auth_object).any? { |s| s.to_sym == str.to_sym }
```
seanlinsley: We can assume that most people use symbols instead of strings, so this prevents 13 KB of string allocations without sacrificing backward compatibility.

avit: Depending on the ruby version, this could cause a memory leak from different string inputs being added to the symbol table.

seanlinsley: That's Ruby 1.8 and below if I recall correctly, and Ransack currently requires 1.9 or above. (However note that on another line I'm using Ruby 2 keyword arguments).

ernie: I want to say it was in the 2.1 series, too, but I could be wrong.

avit: http://www.cvedetails.com/cve/CVE-2013-1854/

avit: Yes, Symbol GC was added in ruby 2.2. This may be OK now, depending on what Ransack wants to support.

lib/ransack/adapters/active_record/base.rb
seanlinsley: It appears that we can use the `to_prepare` Rails callback to expire the cache, though that would require that we do one of these:
- instead of using instance variables, cache the values in a global object
- cache the values in instance variables, but via a module that would hook into `to_prepare`

colesteaks:
@seanlinsley any updates here?

seanlinsley:
I haven't focused on this in quite a while. If you're interested in contributing @colesteaks, I'd be happy to pair with you on this sometime.

lib/ransack/context.rb
seanlinsley: ref #857 CC @seanfcarroll

lib/ransack/adapters/active_record/base.rb
seanlinsley: It appears that we don't currently have tests to verify that Rails code reloading works with Ransack. IIRC ActiveAdmin has tests of that type, but they require a lot of setup code.

seanlinsley: Actually `to_prepare` is unnecessary since these methods are defined on the models themselves. In the Paper Trail case, adding `has_paper_trail` to the model would unload & reload the model, so `ransackable_associations` is no longer cached.

AFAICT there aren't any scenarios where this could cause a bug.

seanlinsley:
I'm going to start using this in my app to see if any bugs appear.

seanfcarroll:
Happy to merge this into master if it's got consensus. Thoughts?

seanlinsley:
My only concern is that the test suite might not have caught the type of bugs that real Rails apps would surface.

lib/ransack/predicate.rb
```diff
@@ -9,34 +9,26 @@ def names
         Ransack.predicates.keys
       end
 
-      def names_by_decreasing_length
-        names.sort { |a, b| b.length <=> a.length }
-      end
```
seanlinsley: ActiveAdmin relies on this method:
```ruby
NoMethodError: undefined method 'names_by_decreasing_length' for Ransack::Predicate:Class

active_admin/filters/humanized.rb in predicates at line 50

          end

          def predicates
            Ransack::Predicate.names_by_decreasing_length
          end

          def ransack_predicate_translation

active_admin/filters/humanized.rb in current_predicate at line 46

          end

          def current_predicate
            @current_predicate ||= predicates.detect { |p| @body.end_with?("_#{p}") }
          end

          def predicates

active_admin/filters/humanized.rb in ransack_predicate_translation at line 54

          end

          def ransack_predicate_translation
            I18n.t("ransack.predicates.#{current_predicate}")
          end

          def active_admin_predicate_translation

active_admin/filters/humanized.rb in body at line 17

          end

          def body
            predicate = ransack_predicate_translation

            if current_predicate.nil?
              predicate = @body.titleize

active_admin/filters/resource_extension.rb in block (6 levels) in search_status_section at line 170

                    else
                      active.filters.each do |filter|
                        li do
                          span filter.body
                          b filter.value
                        end
                      end
```

seanlinsley: It seems like ActiveAdmin was duplicating behavior that Ransack already provided, so it could be simplified to:
```ruby
def current_predicate
  Ransack::Predicate.detect_from_string @body
end

seanlinsley: Ah, https://github.com/activeadmin/activeadmin/pull/4951 replaced this code with `Ransack::Translate.predicate`

seanlinsley:
After also upgrading ActiveAdmin from 1.0 to 1.2, I'm now using this patch in production.

seanfcarroll:
I vote we merge it. I'll do so over the weekend unless anyone disagrees.

seanlinsley:
We haven't seen any issues so far 👍

seanlinsley:
Still no issues, so this should be good to go.

adet4ever:
I was about to open a issue before I found this: 
Attempts upgrade active admin to 1.2 and hoping, I should be good. However Active admin 1.2 requires ruby 2.2 

The coincidence that dev works from 2 days ago. 
Here are the things I have typed up:


XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXx
### Steps to reproduce
Upgrade in dev to rails 4.2.7.1 from rails 4.2.1 was two days ago
Upgrade in prod to rails 4.2.7.1 from rails 4.2.1 was yesterday
I upgraded my rails version from 4.2.1 to 4.2.7.1 in development and everything seems to work fine.

When I tested in a prod environment, In Active admin,  when I try to do a text filter, I'm getting a error

    1: insert_tag renderer_for(:index)
  /home/deploy/.rvm/gems/ruby-2.1.3/bundler/gems/active_admin-dce083189c46/lib/active_admin/filters/humanized.rb:50:in `predicates'
  /home/deploy/.rvm/gems/ruby-2.1.3/bundler/gems/active_admin-dce083189c46/lib/active_admin/filters/humanized.rb:46:in `current_predicate'
  /home/deploy/.rvm/gems/ruby-2.1.3/bundler/gems/active_admin-dce083189c46/lib/active_admin/filters/humanized.rb:54:in `ransack_predicate_translation'
  /home/deploy/.rvm/gems/ruby-2.1.3/bundler/gems/active_admin-dce083189c46/lib/active_admin/filters/humanized.rb:17:in `body'
....... 

### Expected behavior
I expected the filter to just work as this works in my dev environment
### Actual behavior
With the NGINX passenger setup, the page just errors out, with the error above in log file
### System configuration
**Rails version**: 4.2.7.1
**Ruby version**: 2.1.3p242

Any suggestion  the ruby upgrade is a must?

seb-sykio:
hi
I'm still having this issue "exception : undefined method `names_by_decreasing_length' for Ransack::Predicate:Class"
I'm using ransack (2.1.0) and activeadmin (1.4.0) 

thanks for help

PR #823, opened 2017-08-03
Cast Postgres money type to float

seanlinsley:
Postgres otherwise handles the string that's passed down, except in the case that the user accidentally enters a string meant for a different attribute entirely. `to_f` converts it into zero, making it a valid database request.

```ruby
>> Table.search(column_eq: '1.5').result.count
# SELECT COUNT(*) FROM "tables" WHERE "tables"."column" = '1.5'

>> Table.search(column_eq: 'foo').result.count
# SELECT COUNT(*) FROM "tables" WHERE "tables"."column" = '0.0'

>> Table.search(column_eq: '100').result.count
# SELECT COUNT(*) FROM "tables" WHERE "tables"."column" = '100.0'
```

### DIFF
--- a/lib/ransack/nodes/value.rb
+++ b/lib/ransack/nodes/value.rb
@@ -36,6 +36,8 @@ module Ransack
           cast_to_float(value)
         when :decimal
           cast_to_decimal(value)
+        when :money
+          cast_to_money(value)
         else
           cast_to_string(value)
         end
@@ -95,6 +97,10 @@ module Ransack
        else
          val.to_s.to_d
        end
+      end
+      
+      def cast_to_money(val)
+        val.blank? ? nil : val.to_f.to_s
       end
 
       def inspect
--- a/lib/ransack/nodes/value.rb
+++ b/lib/ransack/nodes/value.rb
@@ -23,87 +23,87 @@ module Ransack
       end
 
       def cast(type)
-         case type
-         when :date
-           cast_to_date(value)
-         when :datetime, :timestamp, :time
-           cast_to_time(value)
-         when :boolean
-           cast_to_boolean(value)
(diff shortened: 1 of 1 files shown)


seanlinsley:
Note that these are the only actual changes in this PR:

screen shot 2017-08-08 at 4 08 44 pm

The other changes are indentation fixes.

lib/ransack/nodes/value.rb
```diff
@@ -23,87 +23,93 @@ def hash
...
+         val
+       elsif val.respond_to?(:to_d)
+         val.to_d
+       else
+         val.to_s.to_d
        end
+      end
+      
+      def cast_to_money(val)
+        val.blank? ? nil : val.to_f.to_s
```
avit: Suggest decimal `to_d` instead of float?

seanlinsley: Are there particular benefits / drawbacks for this scenario?

avit: Possibly issues with rounding: not sure how this value is used downstream.

seanlinsley: I don't imagine any rounding would occur; it's a string that's coerced into a number, then back into a string.

>not sure how this value is used downstream

It's just used in the resulting SQL query, AFAICT.

avit: Ok. Just pointing it out since "money" and "float" in the same sentence always picks up on my bug antennas. 😄

seanfcarroll: So @avit is this good to go?

seanlinsley: It should be. We've been using it in production for months without issue.

PR #830, opened 2017-09-14
Disable ActionView::Helpers::Tags monkey patch on Rails 5

ybart:
https://github.com/rails/rails/commit/530b5fea8f3e04c45a20ff3dcc8d473a6d25c1b4#diff-1b43bc32bcc4987e19cfb3ba2c1c7733

### DIFF
--- a/lib/ransack/helpers/form_builder.rb
+++ b/lib/ransack/helpers/form_builder.rb
@@ -1,13 +1,15 @@
 require 'action_view'
 
-module ActionView::Helpers::Tags
-  # TODO: Find a better way to solve this issue!
-  # This patch is needed since this Rails commit:
-  # https://github.com/rails/rails/commit/c1a118a
-  class Base
-    private
-    def value(object)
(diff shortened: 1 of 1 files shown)


al2o3cr:
We've run into the same issue trying out Rails master (yay ActiveStorage).

It would be more in keeping with the way version-switching is done to check `ActionView::VERSION::STRING` - `Rails` may not have been loaded yet. 

The regex check also hits the wrong versions - it removes the patch for 5.0 and 5.1 which need it, and adds it back for Rails >= 6. The codebase standard would be something like `::ActionView::VERSION::STRING >= Constants::RAILS_5_2`.

ybart:
I agree. This has been done quickly. I will need to make this cleanly.

seanfcarroll:
@ybart I'm thinking this is no longer an issue and this PR can be closed. What do you think?

seanfcarroll:
Closing as unneeded.

PR #841, opened 2017-11-01
Bug: negative condition with a default scope

janklimo:
Not a full-fledged PR, but a starting point for fixing a bug that occurs when `default_scope` is present on the joined association (see the regression spec below).

screen shot 2017-11-01 at 12 50 03 pm

I am not overly familiar with `ransack`'s internals, so I was hoping you guys would know how to handle this case. @avit you did some nice work on https://github.com/activerecord-hackery/ransack/commit/96d0fd75ade6a6ca019ccc03cca6d876f0e12ecf and this is directly related to your changes, so I was wondering if you have a fix in mind?

### DIFF
--- a/lib/ransack/adapters/active_record/context.rb
+++ b/lib/ransack/adapters/active_record/context.rb
@@ -184,7 +184,7 @@ module Ransack
           join_table = join_root.left
           correlated_key = join_root.right.expr.left
           subquery = Arel::SelectManager.new(association.base_klass)
-          subquery.from(join_root.left)
+          subquery.from(join_table)
           subquery.project(correlated_key)
           join_constraints.each do |j|
             subquery.join_sources << Arel::Nodes::InnerJoin.new(j.left, j.right)
--- a/spec/ransack/helpers/form_helper_spec.rb
+++ b/spec/ransack/helpers/form_helper_spec.rb
@@ -849,6 +849,11 @@ module Ransack
         before do
           Ransack.configure { |c| c.search_key = :example }
         end
+
+        after do
+          Ransack.configure { |c| c.search_key = :q }
+        end
+
         subject {
           @controller.view_context
           .search_form_for(Person.search) { |f| f.text_field :name_eq }
--- a/spec/support/schema.rb
+++ b/spec/support/schema.rb
@@ -157,12 +157,6 @@ module Namespace
   end
 end
 
-module Namespace
-  class Article < ::Article
-
-  end
-end
-
 class Comment < ActiveRecord::Base
   belongs_to :article
   belongs_to :person
--- a/spec/ransack/adapters/active_record/base_spec.rb
+++ b/spec/ransack/adapters/active_record/base_spec.rb
@@ -102,6 +102,15 @@ module Ransack
           end
         end
 
+        context 'negative condition where the joined association has a default scope' do
+          it 'applies the default scope' do
+            # Article has default_scope
+            s = Person.ransack(articles_title_not_eq: 'Breaking news')
+            sql = s.result.to_sql
+            expect(sql).to include "WHERE ('default_scope' = 'default_scope')"
+          end
+        end
+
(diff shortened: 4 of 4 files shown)


lib/ransack/adapters/active_record/context.rb
```diff
@@ -181,7 +181,6 @@ def remove_association(association)
         def build_correlated_subquery(association)
           join_constraints = extract_joins(association)
           join_root = join_constraints.shift
-          join_table = join_root.left
```
janklimo: this was unused

spec/ransack/helpers/form_helper_spec.rb
```diff
@@ -849,6 +849,11 @@ module Helpers
         before do
           Ransack.configure { |c| c.search_key = :example }
         end
+
+        after do
+          Ransack.configure { |c| c.search_key = :q }
```
janklimo: Specs are really unstable for me locally. To me it looks like the configuration options and their modifications leak across specs and result in intermittent failures depending on the order in which specs are run. This is a good example: unless we return the option value back to its default, it will affect other specs.

Do you guys see that behavior locally or am I missing something?

spec/support/schema.rb
```diff
@@ -157,12 +157,6 @@ class Article < ::Article
   end
 end
 
-module Namespace
-  class Article < ::Article
-
-  end
-end
-
```
janklimo: this is a duplicate of the lines above it

avit:
My colleague is working on something coincidentally around this issue right now! 😄 The fix goes a little bit deeper than this... Expect an update soon!

janklimo:
Great news @avit 👏   any updates?

avit:
@janklimo some good progress, but no update or pull-request just yet.

janklimo:
@avit have you guys made progress here? Wondering if I should deploy a workaround for this or rather wait for a new release?

spec/ransack/helpers/form_helper_spec.rb
seanlinsley: It's been a while since I ran the test suite myself, but if it's being run in a random order, `rspec bisect` can help identify the tests that are causing other tests to fail.

stas:
I'm not sure what's `default_scope`, but I'm using the patch below to fix the two related issues introduced here:
https://github.com/activerecord-hackery/ransack/commit/96d0fd75ade6a6ca019ccc03cca6d876f0e12ecf

The first one is related (maybe connected to this one) to the sub-queries being split and negated and next, the condition being negated again.

Consider this (where a tag has many tag_categories):
```ruby
class Tagging < ActiveRecord::Base
  belongs_to :tag
  has_many :tag_categories, through: :tag
end
```

A query like this: 
```ruby
Tagging.ransack(tag_category_title_not_eq: 'Test').result.to_sql
```

Renders a query like this:
```sql
SELECT "taggings".*
 FROM "taggings"
 WHERE (

--- The tagging ID tries to match the tag ID below...

   "taggings"."id" NOT IN (
       SELECT "tags"."id"
       FROM "tags"
       INNER JOIN "tag_category_tags" ON
         "tag_category_tags"."tag_id" = "tags"."id"
       INNER JOIN "tag_categories" ON
         "tag_categories"."id" =
           "tag_category_tags"."tag_category_id"
       WHERE "tags"."id" = "taggings"."id"

--- The double negation invalidates the clause below...

         AND NOT ("tag_categories"."title" != 'Test')))
```

Reverting the change-set fixes the issue:
```ruby
# Drop this into an initializer...
Ransack::Nodes::Condition.class_eval do
  alias_method :original_arel_predicate, :arel_predicate

  def arel_predicate
    attributes.map { |att| format_predicate(att) }.reduce(combinator_method)
  end
end
```

@janklimo could you give it a try and see if it help in any way? Unfortunately I'm a bit overwhelmed at the moment, I wish I could set up a proper PR though :see_no_evil:

PR #848, opened 2017-12-05
Fix custom search with scope

imxiaozhi:
Example:
```ruby
class Shoot::Video < ApplicationRecord
  acts_as_paranoid

  scope :choiced, -> (boolean = true) {
    if boolean
      where.not(choiced_at: nil)
    else
      where(choiced_at: nil)
    end
  }

  def self.ransackable_scopes(auth_object = nil)
    %i(choiced)
  end
end
```
when false be passed as boolean to scope, here is no scope search.
```ruby
Shoot::Video.ransack(choiced: 'false', s: 'hot desc').result.to_sql
# SELECT `shoot_videos`.* FROM `shoot_videos` WHERE `shoot_videos`.`deleted_at` IS NULL ORDER BY `shoot_videos`.`hot` DESC
```

### DIFF
--- a/lib/ransack/context.rb
+++ b/lib/ransack/context.rb
@@ -46,7 +46,11 @@ module Ransack
 
     def chain_scope(scope, args)
       return unless @klass.method(scope)
-      @object = @object.public_send(scope, *args)
+      @object = if scope_arity(scope) < 1 && args == true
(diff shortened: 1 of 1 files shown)


imxiaozhi:
It`s my mistake.
I should use choiced and unchoiced scope.

PR #854, opened 2017-12-15
nil shoud be false

bilalbash:


seanfcarroll:
Hi @bilalbash I think nil should remain as 'nil' and not transform to false. This would translate to NULL at the database level.

bilalbash:
thanks @seanfcarroll rather using false check i used null check to fix the problem i faced locally. because i wanted all null record as well as false ones.

Issue #835, opened 2017-10-09
Turkish Translations wanted


seanfcarroll:
If anyone speaks Turkish, we are missing Turkish translations, and would appreciate a PR with them.

Teşekkür ederim :-)

puskuruk:
Then i can start to translation! 

Bir şey değil :)

seanfcarroll:
Closed https://github.com/activerecord-hackery/ransack/pull/856

PR #856, opened 2017-12-22
Add turkish translations

sadikay:
Turkish translation for #835.


seanfcarroll:
Thanks for the translations @sadikay !

When I have time I'll look into why the tests are failing.

seanfcarroll:
I think I see the problem:

```
en:
   ransack:
     search: "ara"
```

Should be: 

```
tr:
   ransack:
     search: "ara"
```

Issue #857, opened 2017-12-29
Drop older Ruby and Rails support


seanfcarroll:
Bring Ransack into line with Polyamorous and drop some older versions.

https://github.com/activerecord-hackery/polyamorous/blob/master/.travis.yml

Posted here in case before proceeding in case this breaks other code, eg ActiveAdmin https://github.com/activeadmin/activeadmin/issues/5254  cc @varyonic

seanlinsley:
I'm always a fan of reducing the complexity of libraries by only supporting the latest major versions of co-dependencies; it's not worth the extra burden it puts on maintainers. If people on old Rails versions want a new feature, they can patch it into their own fork or convince a maintainer to backport it.

Though I would recommend an X or Y version bump (of X.Y.Z) so that it's easy to release bug fixes to older versions if necessary.

varyonic:
Ruby 2.2 is [EOL end of this March, 2018](https://www.ruby-lang.org/en/downloads/branches/).
Rails 4.2 will receive severe security fixes [until 6.0 is available](https://stackoverflow.com/questions/43050213/ruby-on-rails-4-2-end-of-life).
It would be nice to match these dates for library maintenance, but not required if it is indeed a burden.

avit:
+1 on this to simplify everything. I had suggested moving the multiple version support around in #416 but never got back around to working on it.

I think it could be structured it so the *current* supported version is the explicit one, but there can still be some facility to include *backports* as separate modules if anyone wants to contribute them.

Personally I still depend on ransack with a Rails 3.2 application (I know, ouch), but can't expect the community to continue supporting that!

seanfcarroll:
@varyonic that makes sense.

seanfcarroll:
So do we go forward with :

#### Rails
5.1, 5.0 and 4.2

As per [this page](http://guides.rubyonrails.org/maintenance_policy.html#unsupported-release-series)

#### Ruby 
2.2 until end of March 2018
2.3
2.4
 
As per [this page](https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/)

PR #858, opened 2017-12-30
Bump Polyamorous to 1.3.2 for Rails 5.2 support

seanfcarroll:
Please add comments / review before this is merged.

### DIFF
--- a/ransack.gemspec
+++ b/ransack.gemspec
@@ -6,8 +6,8 @@ Gem::Specification.new do |s|
   s.name        = "ransack"
   s.version     = Ransack::VERSION
   s.platform    = Gem::Platform::RUBY
-  s.authors     = ["Ernie Miller", "Ryan Bigg", "Jon Atack"]
-  s.email       = ["ernie@erniemiller.org", "radarlistener@gmail.com", "jonnyatack@gmail.com"]
+  s.authors     = ["Ernie Miller", "Ryan Bigg", "Jon Atack","Sean Carroll"]
+  s.email       = ["ernie@erniemiller.org", "radarlistener@gmail.com", "jonnyatack@gmail.com","sfcarroll@gmail.com"]
   s.homepage    = "https://github.com/activerecord-hackery/ransack"
   s.summary     = %q{Object-based searching for Active Record and Mongoid (currently).}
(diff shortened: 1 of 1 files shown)


ransack.gemspec
```diff
@@ -6,8 +6,8 @@ Gem::Specification.new do |s|
   s.name        = "ransack"
   s.version     = Ransack::VERSION
   s.platform    = Gem::Platform::RUBY
-  s.authors     = ["Ernie Miller", "Ryan Bigg", "Jon Atack"]
-  s.email       = ["ernie@erniemiller.org", "radarlistener@gmail.com", "jonnyatack@gmail.com"]
```
seanlinsley: My understanding of `authors` is that it represents the people who wrote a substantial portion of the library. I could certainly be wrong, though.

On the other extreme, ActiveAdmin only lists Greg Bell as an author:

screen shot 2017-12-31 at 2 58 10 pm

seanfcarroll: Any chance the code could be tested @seanlinsley ?

bbonislawski: Agreed. Do not change authors. I wonder even if it's required to change anything since ~1.3 means 1.3.2 too

seanlinsley: Unless there's a reason to prevent applications from using 1.3.0 and 1.3.1, @bbonislawski appears to be right.

ernie: Given there's no "maintainers" section, I'm fine with adding an author, personally.

seanfcarroll: Thanks @ernie

seanfcarroll:
@bbonislawski it was added as 1.3.2 as it wasn't picking up the changes in Polyamorous. Perhaps merge this change with https://github.com/activerecord-hackery/ransack/pull/859 and we can try for a 5.2 - friendly release.

I think we should also consider putting Ransack on a bit of a diet, and dropping older Rails support. https://github.com/activerecord-hackery/ransack/issues/857 what do you think?

seanlinsley:
👍 (I posted my thoughts on #857)

seanfcarroll:
OK to merge this?

seanlinsley:
Looks good to me

PR #866, opened 2018-02-02
Let escape_wildcards not escape '|'

amatsuda:
The regexp `/([\\|\%|_|.])/` means `\` or `|` or `%` or `|` or `_` or `|` or `.`, which includes unwanted `|` character multiple times.


seanfcarroll:
Thanks @amatotsuji, I'll merge this, but it would be good to have specific tests written for these types of PRs. Any chance to add some tests?

PR #859, opened 2018-01-02
add support for rails 5.2

bbonislawski:

Issue #849, opened 2017-12-09
ActiveRecord 5.2 Support


bbonislawski:
Did someone succesfully run ransack with AR 5.2? Build is failing on rails master branch(actually it's failing everywhere). Is someone working on that?

halo:
It's an underlying issue [over here](https://github.com/activerecord-hackery/polyamorous/issues/26).

pioz:
Also in `lib/ransack/adapters/active_record/context.rb:282` pass 3 args to `JoinDependency.new` but with Rails 5.2 seems the constructor take 4.

halo:
@pioz Haha, I just ended up in the same rabbit hole as you. You win :) Let me link to the linkes:

3 arguments passed here:
https://github.com/activerecord-hackery/ransack/blob/724d020442dad60a1a6d9e00bbae7681350c9fb1/lib/ransack/adapters/active_record/context.rb#L282-L284

4 required here:

https://github.com/rails/rails/blob/38b2904f2c9d89868e96edd71c5f3143ca235a88/activerecord/lib/active_record/associations/join_dependency.rb#L91

halo:
Also, this override in ransack accepts an argument:

https://github.com/activerecord-hackery/ransack/blob/724d020442dad60a1a6d9e00bbae7681350c9fb1/lib/ransack/helpers/form_builder.rb#L9

But the original in Rails no longer takes an argument:

https://github.com/rails/rails/blob/38b2904f2c9d89868e96edd71c5f3143ca235a88/actionview/lib/action_view/helpers/tags/base.rb#L38

pioz:
Also in the `join_root` attr used in https://github.com/activerecord-hackery/ransack/blob/724d020442dad60a1a6d9e00bbae7681350c9fb1/lib/ransack/adapters/active_record/ransack/context.rb#L34
is become protected:

https://github.com/rails/rails/blob/c5ab6e51a7b9ee05a2d262a72c7130b9c1d1b0ce/activerecord/lib/active_record/associations/join_dependency.rb#L159

bbonislawski:
So I have working version for me:
https://github.com/spark-solutions/polyamorous
and
https://github.com/spark-solutions/ransack

seanfcarroll:
https://github.com/activerecord-hackery/polyamorous/pull/28

seanfcarroll:
[...]

PR #868, opened 2018-02-03
Add support for Rails 5.2

### LINKED ISSUE #853: Form builder broken in Rails 5.2
Hi all,

On the 5.2.0beta branch of rails this line

https://github.com/activerecord-hackery/ransack/blob/724d020442dad60a1a6d9e00bbae7681350c9fb1/lib/ransack/helpers/form_builder.rb#L9

fails as value isn't called with any arguments from rails tag methods, eg:

https://github.com/rails/rails/blob/659c516bef2781cc66865fc78ed5dce682566d26/actionview/lib/action_view/helpers/tags/base.rb#L53

object is available here, so shouldn't need to be passed as an arg:

https://github.com/rails/rails/blob/659c516bef2781cc66865fc78ed5dce682566d26/actionview/lib/action_view/helpers/tags/base.rb#L10

so the change should be easy, but I can't figure out how to bundle ransack against rails 5.2.0beta to fix the specs. Any guidance and I'd be happy to help out with a PR :grin:

varyonic:
Fixes #853. Closes #859, #849.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,10 @@ rvm:
   - 2.0
 
 env:
+  - RAILS=5-2-stable DB=sqlite3
+  - RAILS=5-2-stable DB=mysql
+  - RAILS=5-2-stable DB=postgres
+
   - RAILS=5-0-stable DB=sqlite3
   - RAILS=5-0-stable DB=mysql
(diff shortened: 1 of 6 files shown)


seanfcarroll:
Thank you @varyonic , will try to get this merged today.

PR #873, opened 2018-02-06
Fix build_joins on Rails 5.2.rc1

glebm:
Fixes the following error:

    protected method `alias_tracker' called for #<ActiveRecord::Associations::JoinDependency:0x0000000002fc3bc0>

The method is still protected in Rails 5.2 and so must be called via `send`.

### DIFF
--- a/lib/ransack/adapters/active_record/context.rb
+++ b/lib/ransack/adapters/active_record/context.rb
@@ -290,7 +290,7 @@ module Ransack
             alias_tracker = ::ActiveRecord::Associations::AliasTracker.create(::ActiveRecord::Base.connection, relation.table.name, join_list)
             join_dependency = JoinDependency.new(relation.klass, relation.table, association_joins, alias_tracker)
             join_nodes.each do |join|
-              join_dependency.alias_tracker.aliases[join.left.name.downcase] = 1
+              join_dependency.send(:alias_tracker).aliases[join.left.name.downcase] = 1
             end
           end
 


seanfcarroll:
Awesome @glebm thanks for the fix. I think there might be some issues see here: https://github.com/activerecord-hackery/ransack/pull/874

It would be super cool if someone could write a test targeting this code. If not possible I'll try to find time this week.

glebm:
Perhaps merge first and the test can come later because this at least fixes the issue?

halo:
I can confirm this bug exists when using Rails 5.2, [here is the protected method](https://github.com/rails/rails/blob/e4719ec047a529dc0759336136f9df18370c1b34/activerecord/lib/active_record/associations/join_dependency.rb#L159-L160).

I can furthermore confirm that this PR fixed it 😎

Thank you @glebm and @bbonislawski just waiting for the rubygems release ;)

seanfcarroll:
@halo I'm waiting before cutting a new gem as we just released v1.8.7 3 days ago. I suspect several of the failed PRs will now be OK, since we have dropped Ruby 2 and older Rails versions. If any of those are ready we can add them into 1.8.8 also.

halo:
@seanfcarroll Alas, I hoped you would not feel stressed by my comment, don't worry :) Since this has been merged into master, I'm fine. Pointing the Gemfile to this canonical repository is considered painless compared to forking :D

Thank you for maintaining this.

seanfcarroll:
@halo, no problem! 

It **was** a lot of work to get this repo cleaned up, but there has been a **lot of help** from many people, including yourself. This is how open-source works, and in particular the PRs by @varyonic for 5.2 and updating supporting versions made it possible to merge other old PRs.

I think Ransack is an amazing gem, and it's worth putting in time to keep it going. It did look a bit shaky at one point.

Thank you for your kind comments. :-)

I think we will wait a week or so, then cut a new version of the gem.

PR #875, opened 2018-02-07
Update build matrix

### LINKED ISSUE #857: Drop older Ruby and Rails support

varyonic:
Closes #857 FWIW.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,8 @@ language: ruby
 sudo: false
 
 rvm:
-  - 2.3.3
-  - 2.2.6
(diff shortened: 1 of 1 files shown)


seanfcarroll:
Awesome. Thanks @varyonic

PR #887, opened 2018-02-28
Fixed missing translation

robinvdijk:


robinvdijk:
Oh I didn't even see that, I just noticed because I used the gteq option. I'll fix it as well!

seanfcarroll:
Great work! Thanks @robinvdijk

PR #893, opened 2018-03-15
Fix connection issue of multiple database

### LINKED ISSUE #881: Ransack no longer works with Secondbase gem.
I opened an issue on Secondbase here: https://github.com/customink/secondbase/issues/56.

It looks like moving from `@engine.connection.schema_cache` to `::ActiveRecord::Base.connection.schema_cache` will exclude tables provided in a secondary schema.

I was using Ransack with ActiveAdmin and Secondbase (to have an admin interface for my Rails app in another app, but share some parts of a database).
Ruby 2.5.0
Rails 5.1.4
Ransack 1.8.7

I understand if this is may not be an urgent/relevant issue. 
The commit that added the change is
https://github.com/activerecord-hackery/ransack/commit/d44575506ccc7cb00e618256707ed53732a07bca. 
I would love to understand what changes in Rails 5.2 it is adapting to, so that I can try submit another solution that will allow gems such as Secondbase to continue to work with Ransack.

If I have missed anything, please let me know and I am happy to add it.

yujideveloper:
This pr fixes #881.

### DIFF
--- a/lib/ransack/adapters/active_record/context.rb
+++ b/lib/ransack/adapters/active_record/context.rb
@@ -27,7 +27,7 @@ module Ransack
           return nil unless attr && attr.valid?
           name         = attr.arel_attribute.name.to_s
           table        = attr.arel_attribute.relation.table_name
-          schema_cache = ::ActiveRecord::Base.connection.schema_cache
+          schema_cache = self.klass.connection.schema_cache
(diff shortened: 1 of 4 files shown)


seanfcarroll:
Awesome! Will test and merge asap.

seanfcarroll:
Nice approach @yujideveloper. Thank you for your contribution.

PR #899, opened 2018-03-22
Doc for distinct improve

atilla777:
It is seems I have found another solution to solve problem with Postgresql and distinct: true.


seanfcarroll:
Thank you @atilla777, documentation PR's are especially welcome!

PR #901, opened 2018-04-09
Fix activerecord 'loading'

madAle:
Checking for ActiveRecord's version throws a missing constant error when it's not defined, e.g. when using mongoid other than AR.
The fix consists in checking if AR is defined.


seanfcarroll:
Thanks for the PR @madAle

Issue #905, opened 2018-04-16
Azerbaijani translations


seanfcarroll:
In order to align the supported translations with ActiveAdmin, we are looking for a community member to provide **az** translations.

seanfcarroll:
Thanks @orkhan

orkhan:
@seanfcarroll u r welcome 👍

PR #913, opened 2018-04-16
Create az.yml

orkhan:
#905


seanfcarroll:
Thank you for the very quick response @orkhan !

PR #915, opened 2018-05-03
Refactor conditional

shioyama:
Just browsing the code I noticed this odd-looking conditional. `detect` returns nil if no result is found, so no need to check it, just return it, right?


shioyama:
Tests are green, so indeed this would seem to be four lines ransack doesn't need :wink:

seanfcarroll:
@shioyama thank you for your contribution. Great to have people go over the codebase and clean it up!

PR #919, opened 2018-05-11
Refactor inject

shioyama:
I noticed a few places with the pattern:

```ruby
return nil unless nodes.size > 0

if nodes.size > 1
  nodes.inject(&:and)
else
  nodes.first
end
```

This can be simplfied to just:

```ruby
nodes.inject(&:and)
```

since `inject` returns the first element if the array only has one element, and `nil` if the array has no elements (which is exactly what the more verbose code is doing).


seanfcarroll:
Thank you @shioyama !

PR #927, opened 2018-06-23
absorb polyamorous

gregmolnar:
As it is in the readme, the next major release should absorb polyamorous. 
/cc @seanfcarroll

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,10 @@ env:
   - RAILS=5-2-stable DB=mysql
   - RAILS=5-2-stable DB=postgres
 
+  - RAILS=v5.2.0 DB=sqlite3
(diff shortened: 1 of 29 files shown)


gregmolnar:
The CI failure is caused by this: https://github.com/rails/rails/commit/50036e673b3b6ff41476814b212b55fabac3a638?diff=unified
Fixing it soon.

seanfcarroll:
Great work @gregmolnar ! The CI is still running which is strange, but I will check in later and merge it if all green.

gregmolnar:
@seanfcarroll I just pushed a commit, that's why it is running again. There is a breaking change on Rails 5-2-stable unfortunately, so I changed the travis config to test agains 5.2.0 and allow failures on 5-2-stable.

PR #928, opened 2018-06-25
move mongoid support to a separate gem

gregmolnar:


seanfcarroll:
Great work @gregmolnar !

PR #929, opened 2018-07-02
:scissors: old rails support

gregmolnar:
We are dropping support for Rails <5. There are 2 skipped specs I want to fix before merging this.


gregmolnar:
@seanfcarroll Could you please give this a quick glance when you have a chance? There is still a skipped spec, but it seems to require some time to debug that issue, so I would rather have this merged and get back to that later.

Issue #641, opened 2016-01-14
"Present" predicate fails with numeric attributes


ialiendeg:
Using "present" predicate with a numeric field returns an empty array, at least with postgresql database. It seems that the second condition shouldn't be used.

```
>> SupplierOrder.ransack(work_order_id_present: true).result
  SupplierOrder Load (0.8ms)  SELECT "supplier_orders".* FROM "supplier_orders" WHERE ("supplier_orders"."work_order_id" IS NOT NULL AND "supplier_orders"."work_order_id" != NULL)
=> []
```

I think the sql should be

```
SELECT "supplier_orders".* FROM "supplier_orders" WHERE "supplier_orders"."work_order_id" IS NOT NULL
```

Good work though, this gem is a must :)

jaredbeck:
I'm seeing the same exact problem with a `datetime` column.

`x != null` is a [contradiction](https://en.wikipedia.org/wiki/Contradiction)

jaredbeck:
Ransack builds the `Condition`'s predicate as follows:

```
::Banana.arel_table[:created_at].not_eq_all([nil, '']).to_sql
=> "(`bananas`.`created_at` IS NOT NULL AND `bananas`.`created_at` != NULL)"
```

This implementation is fine for a `varchar`:

```
::Banana.arel_table[:color].not_eq_all([nil, '']).to_sql
=> "(`bananas`.`color` IS NOT NULL AND `bananas`.`color` != '')"
```

As currently implemented, `present` seems only suited for string-like columns.

@seanfcarroll, should we try to fix `present` for non-string columns, or should we just note this limitation in the docs?

gregmolnar:
Let's live with the limitation for now an put this in the backlog.

jaredbeck:
For posterity, I ended up using the `*_not_null` predicate instead of `*_present`.

benlieb:
@jaredbeck *_not_null on a date field gets:

```
 undefined method `eq' for #<Arel::Nodes::Equality:0x00007fd5ee4f1cf0>
2.5.3/gems/ransack-2.1.1/lib/ransack/adapters/active_record/context.rb:173:in `build_correlated_subquery'
```

prwhitehead:
Could I make a suggestion to add something to the docs in the `present` and `blank sections as this was pretty confusing for a while until i found this bug issue.

PR #934, opened 2018-07-20
Docs: Note limitation of the *_present predicate

jaredbeck:
Per https://github.com/activerecord-hackery/ransack/issues/641, `*_present` is only compatible with string-like columns.


gregmolnar:
@jaredbeck Thank you

PR #939, opened 2018-08-07
Do not override `join_constraints`

y-yagi:
In the current master, an error occurs when performing a specific join process in AR 5.2.1(My bad!).

Reproduce script is following.

```
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  gem "rails", "5.2.1.rc1"
  gem "byebug"
  gem "ransack", github: "activerecord-hackery/ransack", require: false
  gem "sqlite3"
  # gem "pg"
end

require "active_record"
require "ransack"
require "minitest/autorun"
require "logger"

ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")

  create_table "set_adopted_books", force: :cascade do |t|
    t.bigint "adopted_book_set_id"
  end

  create_table "adopted_book_sets", force: :cascade do |t|
    t.bigint "library_id"
  end

  create_table "libraries", force: :cascade do |t|
  end
end

class SetAdoptedBook < ActiveRecord::Base
  belongs_to :adopted_book_set
end

class AdoptedBookSet < ActiveRecord::Base
  belongs_to :library
  has_many   :set_adopted_books, dependent: :destroy
end

class Library < ActiveRecord::Base
end

class BugTest < Minitest::Test
  def test_association_stuff
    refute SetAdoptedBook.joins(adopted_book_set: :library).exists?
    # => ActiveRecord::StatementInvalid: SQLite3::SQLException: ambiguous column name: libraries.id: SELECT  1 AS one FROM "set_adopted_books" INNER JOIN "adopted_book_sets" ON "adopted_book_sets"."id" = "set_adopted_books"."adopted_book_set_id" INNER JOIN "libraries" ON "libraries"."id" = "adopted_book_sets"."library_id" INNER JOIN "libraries" ON "libraries"."id" = "adopted_book_sets"."library_id" LIMIT ?
  end
end
```

ransack overrides the join process, and it is an error because there is not enough processing there.
However, it does not seem to be necessary to override the process at the moment. So, I removed the overriding methods.

### DIFF
--- a/lib/polyamorous/activerecord_5.2.1_ruby_2/join_dependency.rb
+++ b/lib/polyamorous/activerecord_5.2.1_ruby_2/join_dependency.rb
@@ -48,63 +48,6 @@ module Polyamorous
       end
     end
 
-    # Replaces ActiveRecord::Associations::JoinDependency#join_constraints
-    #
-    # This internal method was changed in Rails 5.0 by commit
-    # https://github.com/rails/rails/commit/e038975 which added
-    # left_outer_joins (see #make_polyamorous_left_outer_joins below) and added
-    # passing an additional argument, `join_type`, to #join_constraints.
-    #
-    def join_constraints(joins_to_add, join_type, alias_tracker)
-      @alias_tracker = alias_tracker
-
-      construct_tables!(join_root)
-
-      joins = join_root.children.flat_map { |child|
-        if join_type == Arel::Nodes::OuterJoin
-          make_polyamorous_left_outer_joins join_root, child
-        else
-          make_polyamorous_inner_joins join_root, child
-        end
-      }
-
-      joins.concat joins_to_add.flat_map { |oj|
-        construct_tables!(oj.join_root)
-        if join_root.match? oj.join_root
-          walk(join_root, oj.join_root)
-        else
-          make_join_constraints(oj.join_root, join_type)
-        end
-      }
-    end
-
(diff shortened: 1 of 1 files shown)


gregmolnar:
Rails 5.1.2 has been released for 2 days now, so I don't really get why are you the 3rd person who complains about not having a compatible version of Ransack released. It is not like I am waiting for months to cut a release and thanks to @y-yagi master is working fine so please just use that till I get the time to cut a release. 
I maintain this project in my free time, not getting paid for it, so you probably understand that work and my family gets priority over this.

jaredbeck:
Hi Greg, Sorry if it sounded like a complaint. I'm just concerned that a few hundred people will also lose an hour of their lives troubleshooting this. Yes, they can use `master` but they will still have lost that hour (maybe more, if they're not good at debugging gems). I'm not complaining, just wanted to point out that this issue will affect a lot of people. Thanks for your work on this.

gregmolnar:
@jaredbeck 2.0.0 with Active Record 5.2.1 support has been released. Please give it a spin and let us know if you encounter any issues.

jaredbeck:
> @jaredbeck 2.0.0 with Active Record 5.2.1 support has been released. Please give it a spin and let us know if you encounter any issues.

Thanks for the quick release Greg. I have tested ransack 2.0.0 and I only found one issue, which I have done my best to isolate: https://github.com/activerecord-hackery/ransack/issues/947

PR #946, opened 2018-08-10
Remove internal methods which are no longer used

kamipo:
`{find,build}_join_association_respecting_polymorphism` are no longer
used since existed them for Active Record 3.2 and 4.0.


gregmolnar:
@kamipo Thank you!

Issue #947, opened 2018-08-10
Breaking change in 2.0.0 re: joins+order


jaredbeck:
This test passes in rails 5.2.0 + ransack 1.8.8.
This test fails in rails 5.2.1 + ransack 2.0.0.
Is this an intentional breaking change, or a bug?

```ruby
# frozen_string_literal: true

begin
  require "bundler/inline"
rescue ::LoadError => e
  $stderr.puts "Bundler version 1.10 or later is required. Please update your Bundler"
  raise e
end

gemfile(true) do
  source "https://rubygems.org"
  git_source(:github) { |repo| "https://github.com/#{repo}.git" }
  gem "ransack", "2.0.0", require: false
  gem "rails", "5.2.1", require: false
  gem "sqlite3"
end

require "active_record"
require "ransack"
require "minitest/autorun"
require "logger"

# This connection will do for database-independent bug reports.
::ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
::ActiveRecord::Base.logger = ::Logger.new(STDOUT)

::ActiveRecord::Schema.define do
  create_table :books, force: true do |t|
  end

  create_table :chapters, force: true do |t|
    t.string :name
    t.belongs_to :book
  end

  create_table :pages, force: true do |t|
    t.belongs_to :chapter
  end
end

class Book < ::ActiveRecord::Base
end

class Chapter < ::ActiveRecord::Base
  belongs_to :book
end

class Page < ::ActiveRecord::Base
  belongs_to :chapter
end

class BugTest < ::ActiveSupport::TestCase
  # Test passes in rails 5.2.0 + ransack 1.8.8
  # Test fails in rails 5.2.1 + ransack 2.0.0
  def test_association_stuff
    b = Book.create!
    c = Chapter.create!(book: b)
    Page.create!(chapter: c)
    search = { chapter_book_id_eq: b.id }
    assert_nothing_raised {
      ::Page.joins(:chapter).order('chapter.name').ransack(search).result(distinct: true)
    }
    # NoMethodError: undefined method `[]' for nil:NilClass
    # /Users/jared/.rbenv/versions/2.4.4/lib/ruby/gems/2.4.0/gems/ransack-2.0.0/lib/ransack/nodes/bindable.rb:42:in `get_attribute'
[...]

PR #951, opened 2018-08-16
Don't return association if table is nil

CGA1123:
Returning an association here where `table == nil` will result in a subsequent `undefined method '[]' for nil:NilClass` being thrown in [`get_attribute`](https://github.com/activerecord-hackery/ransack/blob/1bd2f31d5a1c462172bad464bb7c0b15e2074cde/lib/ransack/nodes/bindable.rb#L38-L38) due to [`Ransack::Adapters::ActiveRecord::context#table_for`](https://github.com/activerecord-hackery/ransack/blob/1bd2f31d5a1c462172bad464bb7c0b15e2074cde//Users/christiangregg/dev/ransack/lib/ransack/adapters/active_record/context.rb#L69-L69) returning nil.

Falling back to `build_association` works as expected.

Related #947

### DIFF
--- a/lib/ransack/adapters/active_record/context.rb
+++ b/lib/ransack/adapters/active_record/context.rb
@@ -268,7 +268,7 @@ module Ransack
 
         def find_association(name, parent = @base, klass = nil)
           @join_dependency.instance_variable_get(:@join_root).children.detect do |assoc|
-            assoc.reflection.name == name &&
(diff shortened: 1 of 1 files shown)


gregmolnar:
@CGA1123 Thanks for your contribution. I cherry picked your commit and added a spec as well in another PR so closing this.

PR #954, opened 2018-08-18
Remove `lib/ransack/adapters/active_record/compat.rb` which is no longer needed

kamipo:
It had already been fixed in Arel 3.0.3 and Rails 5.0 requires Arel 7.0.

https://github.com/rails/arel/pull/112


gregmolnar:
thank you @kamipo

PR #981, opened 2018-11-12
Fix calling String#split twice to get segments

gssbzn:
`str.splice` was being called twice in the same method to get the segments, remove one of these calls
Other small codestyle fixes


seanfcarroll:
Looks reasonable, I think this can be merged.

PR #984, opened 2018-11-29
copy polyamorous files for ActiveRecord 5.2.2

paulccarey:
Using Rails 5.2.2 breaks ransack due to a search path dependancy in Polyamorous.

This fix simply copies the existing modules for AR 5.2.1 to fix the load path issue causing by upgrading to Rails 5.2.2


gregmolnar:
@paulccarey Thanks, but it would be better to just load the same set of files if they are working for both 5.2.1 and 5.2.2. I did that in #985 and will merge it if Travis is happy.

paulccarey:
ah perfect, yup I agree much better! 👍

PR #985, opened 2018-11-29
load 5.2.1 polyamorous files for >= 5.2.1

gregmolnar:


kennyadsl:
Hey there! 👋 

I think that this is needed in all gems that depend on both Rails and Ransack and are not locking them to a specific version.

Is there any chance to release a new version of the gem with this change?

gregmolnar:
@kennyadsl I will cut a release tomorrow.

lucianosousa:
why you guys just didn't remove the line 15?

gregmolnar:
@lucianosousa 5.2.0 and >= 5.2.1 requires different files to be loaded.

Issue #852, opened 2017-12-13
Feature suggestion: Warn about unexpected search params


jaredbeck:
I have a `Dog`. Dogs can't fly. They don't have wings. The `dogs` table does not have a `wing_length` column. If I call `Dog.ransack(wing_length_lt: 42)` I'd like ransack to raise an error.

Dogs do have ears. The `dogs` table has an `ears` column. If I call `Dog.ransack(ears_banana: 1)` I'd also like ransack to raise an error (maybe it already does?) because "banana" is not a reasonable question to ask about ears.

I don't think that logging a warning is useful enough, because no one sees warnings. Only errors get seen.

This feature should be off by default, to avoid classification as a breaking change.

I imagine that custom ransackers will complicate this feature.

jaredbeck:
I've been informed about `config.ignore_unknown_conditions` (https://github.com/activerecord-hackery/ransack/wiki/Configuration) and will try that. Closing.

jaredbeck:
`ignore_unknown_conditions` seems to work well, and helped me identify two bugs, and a half-dozen bad tests. Great feature, thanks.

Issue #535, opened 2015-04-29
Introduce ransack! so that invalid filters raise an exception


linkyndy:
It would be very useful to provide a method, `ransack!`, that will raise an exception if, say, a filter is not valid. Here's an example:

```ruby
class Client
  has_one :address
end

class Address
  belongs_to :client
end
```

If you do:

```
Address.ransack('client_name_cont' => 'bla')
```

and `name` is a valid field of `Client`, the above would work. But if you do:

```
Address.ransack('filter_not_existing_cont' => 'bla')
```

an error would be more appropriate to be raisen instead of returning all addresses (`SELECT "addresses".* FROM "addresses"`).

I find `ransack!` to be very useful; what do you think?

jonatack:
Hi @linkyndy, there is a configuration option `ignore_unknown_conditions ` described [here in the wiki](https://github.com/activerecord-hackery/ransack/wiki/Configuration), that when set to false raises if a query contains an unknown predicate or attribute. Do you mean something different?

linkyndy:
Indeed, that's what I was referring to. While it's already there, I believe having a `!` counterpart to `ransack` would be both intuitive and helpful.

jonatack:
Feel free to make a PR :smiley:

PR #990, opened 2018-12-12
Proposal: Don't ignore unknown conditions by default.

oneamtu:
Hi,

Big fan of ransack! We've been using it at my company for a while now. One thing that we bumped into recently is that, by default, if you pass in a malformed parameter to ransack, it will ignore it, and by default return the whole collection. That's proved to be an unsafe way to operate for us, since it opened up clients who use our APIs to the possibility that they'll have a badly named search term, and instead of scoping the data to one person, they'll get a bunch of people's data. We did some digging, and discovered `ignore_unknown_conditions`, and will use that in all our apps. I wanted to see if you all would be open to changing the default behavior, since it seems like having `ignore_unknown_conditions => false` is a safer default than the current one. 

Seems like other people bumped into this at times unexpected behavior in the past.

https://github.com/activerecord-hackery/ransack/issues/852
https://github.com/activerecord-hackery/ransack/issues/535

Either way, thank you maintainers for your continued work on this gem. Cheers!

### DIFF
--- a/lib/ransack/configuration.rb
+++ b/lib/ransack/configuration.rb
@@ -28,7 +28,7 @@ module Ransack
 
     self.options = {
       :search_key => :q,
-      :ignore_unknown_conditions => true,
+      :ignore_unknown_conditions => false,
       :hide_sort_order_indicators => false,
       :up_arrow => '&#9660;'.freeze,
       :down_arrow => '&#9650;'.freeze,


gregmolnar:
I am closing this since all specs are failing, but we will consider it for the next major release.
@seanfcarroll What do you think?

PR #1001, opened 2019-02-07
Support Rails 6 in #sort_link helper

sandrew:
Currently in Rails 6.0.0.beta1 it dies with error:
`ActionController::UnfilteredParameters: unable to convert unpermitted parameters to hash`

This small fix allows it to work correctly

### DIFF
--- a/lib/ransack/helpers/form_helper.rb
+++ b/lib/ransack/helpers/form_helper.rb
@@ -145,7 +145,7 @@ module Ransack
         private
 
           def parameters_hash(params)
-            if ::ActiveRecord::VERSION::MAJOR == 5 && params.respond_to?(:to_unsafe_h)
+            if ::ActiveRecord::VERSION::MAJOR >= 5 && params.respond_to?(:to_unsafe_h)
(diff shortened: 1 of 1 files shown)


lib/ransack/helpers/form_helper.rb
```diff
@@ -145,7 +145,7 @@ def html_options(args)
         private
 
           def parameters_hash(params)
-            if ::ActiveRecord::VERSION::MAJOR == 5 && params.respond_to?(:to_unsafe_h)
+            if (5..6).include?(::ActiveRecord::VERSION::MAJOR) && params.respond_to?(:to_unsafe_h)
```
gregmolnar: I would prefer this condition:

```suggestion
            if ::ActiveRecord::VERSION::MAJOR >= 5 && params.respond_to?(:to_unsafe_h)
```

sandrew: Done

mmzoo:
Thank you @sandrew!

Issue #902, opened 2018-04-13
Regression in polymorphic associations in ActiveRecord 5.2


rzane:
This issue only affects Active Record 5.2.

In Active Record < 5.2, polymorphic joins will include a condition on the `_type` column.

```sql
LEFT OUTER JOIN "people"
ON "people"."id" = "notes"."notable_id" AND "notes"."notable_type" = 'Person'
```

In Active Record 5.2, the condition will not be included:

```sql
LEFT OUTER JOIN "people"
ON "people"."id" = "notes"."notable_id"
```

---

I created an issue on the polyamorous repo, but I figured more users are likely to see the issue here.

Sorry for the noise, here's the original issue: https://github.com/activerecord-hackery/polyamorous/issues/37

seanfcarroll:
No problem re: noise. Is it possible to create some failing tests on a branch?

rzane:
Sure, I'll see if I can come up with a reproduction case through the Ransack API. I use polyamorous for [baby_squeel](https://github.com/rzane/baby_squeel), and that's where I noticed the issue. Though, I'm pretty sure I was able to reproduce using ransack.

Since baby_squeel depended on polyamorous, and since polyamorous was merged into this repo, baby_squeel now has to depend on ransack. In my attempt to make baby_squeel compatible with Active Record 5.2.1, I think I 

https://github.com/activerecord-hackery/ransack/issues/955

dholdren:
I ran into this issue as well. I'm using ransack 2.1.1 which now includes the polyamorous code

gregmolnar:
@rzane Can you verify that this PR fixes your issue? https://github.com/activerecord-hackery/ransack/pull/1004

seanfcarroll:
Closed due to inactivity.

Issue #955, opened 2018-08-20
Rails 5.2.1: Join type for nested relation becomes INNER, not LEFT OUTER


yhatt:
I have tried to upgrade own Rails from 5.2.0 to 5.2.1, but several tests using ransack query to nested relation have failed because of a join type difference.

A simple example to reproduce:

```ruby
class A < ApplicationRecord
  belongs_to :b
end

class B < ApplicationRecord
  has_many :as
  has_many :cs
end

class C < ApplicationRecord
  belongs_to :b
end

# Search from A's related text `bs`.`name` or `cs`.`name`
A.ransack({ b_name_or_b_c_name_cont: 'foobar' }).result.to_sql

# Rails 5.2.0:
# SELECT "as".* FROM "as" LEFT OUTER JOIN "bs" ON "bs"."id" = "as"."b_id" LEFT OUTER JOIN "cs" ON "cs"."b_id" = "bs"."id" WHERE ("bs"."name" LIKE '%foobar%' OR "cs"."name" LIKE '%foobar%')
#                                                                         ^^^^^^^^^^
# Rails 5.2.1:
# SELECT "as".* FROM "as" LEFT OUTER JOIN "bs" ON "bs"."id" = "as"."b_id" INNER JOIN "cs" ON "cs"."b_id" = "bs"."id" WHERE ("bs"."name" LIKE '%foobar%' OR "cs"."name" LIKE '%foobar%')
#                                                                         ^^^^^
```

In Rails 5.2.1 and Ransack 2.0.1, the search query for `B`'s name will never match if `C` was not related to any records from `B`.

I don't know which one is the correct query on Rails 5.2.1, but it is sure that is losing compatible with the previous version.

ozzyaaron:
I find it hard to say anything for certain, BUT this appears to certainly be a bug. In our case we're doing a `_not_null` ransack query and so it cannot use an inner join for this type of query.

igor04:
it looks like with AR 5.2.1 and `INNER JOIN` instead of `LEFT OUTER` there is no way to have valid search results with blank associations, because inner join will reject all results without association no matter that in condition we use `or`, ex:  `own_field_or_association_field_coun` rejects records without  association even if `own_field` matches (works fine with AR 5.2.0)

seanfcarroll:
A PR with some failing tests would be super-helpful on this one.

varyform:
[...]

PR #1004, opened 2019-02-12
fix polymorphic joins with rails > 5.2

### LINKED ISSUE #996: AR 5.2.1 and 5.2.2 transitive polymorphic association results in INNER JOIN which breaks "or" searches
# Problem
When searching a model's polymorphic association's association's field in Rails 5.2 ransack generates a `INNER JOIN` which causes `or` type searches to break
(i.e. a Note's notable Article's comment's body)

## Working combination: Rails 5.0, ransack 2.1.1 (latest):
This generates `LEFT OUTER JOIN`s for everything
```
➜  ransack git:(master) rake console
[1] pry(main)> puts Note.ransack("notable_of_Article_type_comments_body_cont" => "First").result.to_sql
SELECT "notes".* FROM "notes" 
LEFT OUTER JOIN "articles" ON "articles"."id" = "notes"."notable_id" AND "notes"."notable_type" = 'Article' AND ('default_scope' = 'default_scope') 
LEFT OUTER JOIN "comments" ON "comments"."article_id" = "articles"."id" 
WHERE ("comments"."body" LIKE '%First%')
=> nil

[2] pry(main)> puts Note.ransack("notable_of_Person_type_name_or_notable_of_Article_type_comments_body_cont" => "First").result.to_sql
SELECT "notes".* FROM "notes" 
LEFT OUTER JOIN "people" ON "people"."id" = "notes"."notable_id" AND "notes"."notable_type" = 'Person' 
LEFT OUTER JOIN "articles" ON "articles"."id" = "notes"."notable_id" AND "notes"."notable_type" = 'Article' AND ('default_scope' = 'default_scope') 
LEFT OUTER JOIN "comments" ON "comments"."article_id" = "articles"."id" 
WHERE ("people"."name" LIKE '%First%' OR "comments"."body" LIKE '%First%')
=> nil
```

## Broken combination Rails 5.2.2 (also 5.2.1), ransack 2.1.1 (latest)
This generates an `INNER JOIN` for the `comments` table so it ignores any other tables we chose to search
```
ransack git:(master) RAILS=v5.2.2 rake console
[1] pry(main)> puts Note.ransack("notable_of_Article_type_comments_body_cont" => "First").result.to_sql
SELECT "notes".* FROM "notes" 
LEFT OUTER JOIN "articles" ON "articles"."id" = "notes"."notable_id" AND ('default_scope' = 'default_scope') 
INNER JOIN "comments" ON "comments"."article_id" = "articles"."id" 
WHERE "comments"."body" LIKE '%First%'
=> nil

[2] pry(main)> puts Note.ransack("notable_of_Person_type_name_or_notable_of_Article_type_comments_body_cont" => "First").result.to_sql
SELECT "notes".* FROM "notes" 
LEFT OUTER JOIN "people" ON "people"."id" = "notes"."notable_id" 
LEFT OUTER JOIN "articles" ON "articles"."id" = "notes"."notable_id" AND ('default_scope' = 'default_scope') 
INNER JOIN "comments" ON "comments"."article_id" = "articles"."id" 
WHERE ("people"."name" LIKE '%First%' OR "comments"."body" LIKE '%First%')
=> nil
```

## Additional bug
You can see in the Rails 5.2.2 example, that the polymorphic associations `LEFT OUTER JOIN` aren't limited by `notable_type`. This causes other issues. This was reported in https://github.com/activerecord-hackery/ransack/issues/902

gregmolnar:
closes #996 and #902 and #955
I still need to add some specs to cover this.


dholdren:
I was able to confirm that this solves my issue. Thanks again.

rzane:
@gregmolnar, I just ran this against baby_squeel's test suite ([branch here](https://github.com/rzane/baby_squeel/tree/ar-521), [results here](https://gist.github.com/rzane/9a395056cfa816c642b3bfd9ecd1db20)).

It seems like this does fix #902 and #955, so great work. It fixed all of the tests that were specifically related to polymorphism. Those tests had been broken in baby_squeel since Active Record 5.2 initially came out.

However, #955 seems to be in play still.  You'll notice that some of the tests that should say `LEFT JOIN` are instead saying `INNER JOIN`.

In order to get the test suite to even run, I had to make [this change](https://github.com/rzane/baby_squeel/compare/rzane:a71a462...rzane:095ae02#diff-83d7946e8d6ab5d5bcc652544af677a0R37). `JoinAssociation#table` seems to be returning `nil` sometimes. The change seems to fix most of the tests, but I suspect that it might be covering up a larger issue, and it appears to be causing the test failures that are showing the wrong table alias. Maybe this is a clue?

gregmolnar:
@rzane Thanks for the info. I need to find a few hours block when I go over again this and find the root cause.

kieranklaassen:
Thanks, guys for the work on this! I would really appreciate it if you can merge this one @gregmolnar .

kieranklaassen:
@gregmolnar I appreciate your efforts very much!! If you do not have time but could share your idea for the fix I might be able to help out here too. Thanks!

pelletencate:
@gregmolnar I'd love to help out here too. We're holding back on updating Rails because we depend on baby_squeel, and it'd be awesome if I can be of assistance in speeding up this process a bit. If it's easier / less time consuming for you to do a brain-dump here, I'm more than happy to try and give it a go!

gregmolnar:
I need to get into the code to get my head around the problem again to be able to make a brain-dump and I will probably just fix it if I manage to get there.  
On a very high level, the issue is we used to override `ActiveRecord::Associations::JoinDependency#make_left_outer_joins`, but that is either removed on Rails 5.2, or we thought we don't need to override it anymore(I can't remember) but we still need a workaround to turn our joins into a left one. Beware, I am writing this without checking the codebase, but this is roughly where I remember to look where debugged this issue.

gregmolnar:
I finally managed to work on this and I think it is fixed now, although I could come up with specs so far. 
Could you please guys verify that it works for you?
@rzane I ran the specs in your branch and some still fails, but those failures might be unrelated? I don't  know much about baby_squeel, but it looks like the issue happens with some aliasing.

gregmolnar:
@rzane Actually, those alias errors looks like caused by ransack too, and there is a PR to fix that: https://github.com/activerecord-hackery/ransack/pull/1005/files
I will merge this first if I get the feedback of it being ok, then will merge that too.

rzane:
Thanks again for digging into this @gregmolnar.

It seems like once a relation starts to do any outer join, all subsequent joins become outer joins? Take this example:

```ruby
Post.joining { author.outer.comments }
```

In the example above, I would expect `posts -> author` to be an outer join and `author -> comments` to be an inner join. In previous Active Record versions, that works as expected. But, that's not what is happening on 5.2.1. Here's the failing spec:

```
expected:
        SELECT "posts".* FROM "posts"
        LEFT OUTER JOIN "authors" ON "authors"."id" = "posts"."author_id"
        INNER JOIN "comments" ON "comments"."author_id" = "authors"."id"
     got:
        SELECT "posts".* FROM "posts"
        LEFT OUTER JOIN "authors" ON "authors"."id" = "posts"."author_id"
        LEFT OUTER JOIN "comments" ON "comments"."author_id" = "authors"."id"
```

Another issue is the alias detection. Here's a gist of what that looks like in Baby Squeel.

```ruby
def find_alias(associations)
   join_association = find_join_association(associations)
   join_association.table
end
```

Baby Squeel is successfully finding the correct JoinAssociation. On previous versions of Active Record, `#table` would return an `Arel::Table` with the correct name. However, when running against 5.2.1, `#table` is nil. I'm honestly not sure if this part is a polyamorous issue or a baby_squeel issue caused by a change in Active Record.

gregmolnar:
I merged this, since it fixes most of the issues and I want to make master work with Rails 6. I will open a follow up issue for the remaining issue.

PR #1005, opened 2019-02-21
Fix wrong table alias when using nested join.

hiichan:
Fix wrong table alias when using nested join.
only ActiveRecord >= 5.2

#374 

please review the PR


gregmolnar:
@hiichan Thank you! I am closing this in favour of https://github.com/activerecord-hackery/ransack/pull/1004. I cherry-picked your commit into that PR.

PR #973, opened 2018-10-23
Add support for sorting by scopes

benlangfeld:
For when we want to sort by something complex such as the result of a SQL function.

**Credit**: This solution was borrowed from [this gist](https://gist.github.com/laserlemon/e5ce4e0d8bac63f4aa6835605ce3ec21).

PR #1006, opened 2019-02-23
[1.8] Backport support for sorting by scopes

simi:
For when we want to sort by something complex such as the result of a SQL function.

Backport of bdd1c245.

reference https://github.com/activerecord-hackery/ransack/pull/973#issuecomment-459722305


gregmolnar:
Thank you @simi

PR #1044, opened 2019-07-21
Fix for #1042 - tests will not run unless RAILS is defined.

### LINKED ISSUE #1042: Difficulty running tests on repository
According to the ```Contributing to Ransack``` guide, in order to run the tests using postgres, I should do
```
bundle install
DB=pg bundle exec rake spec
```
but when I do that, I get an exception
```
Failure/Error:
  establish_connection(
    adapter: 'sqlite3',
    database: ':memory:'
  )

LoadError:
  Error loading the 'sqlite3' Active Record adapter. Missing a gem it depends on? can't activate sqlite3 (~> 1.4), already activated sqlite3-1.3.13. Make sure all dependencies are added to Gemfile.
```
Similarly for mysql.  Using git bisect, I found the issue starts with this [commit](https://github.com/activerecord-hackery/ransack/commit/131e52aec0e0acd910b9480683fd685ae52264c5).

Looking at your travis file, I found a work around is to do
```
RAILS=v5.2.3 DB=pg bundle install
DB=pg bundle exec rake spec
```

obromios:
This is a pull request to fix #1042. This issue means that
the following commands
```
bundle install
bundle exec rake spec
```
raise an exception when run on a cloned master branch of the repository.

The problem was caused by a change to the default version of Rails. This change meant that the ransack gemspec incorrectly sets the version of sqlite3 when the RAILS env variable is undefined.

The fix was to alter the gemspec to set the correct version of sqlite3.

In order to prevent this issue reoccurring three more tests were added to .travis.yml to make sure the tests pass when the RAILS variable is not set.

There are a couple of issues for the ransack team to consider:

- The version  of sqlite3 specified in the polyamorous gemspec is not dynamically set. This does not seem to matter, but I am not sure that it is OK to leave this to what looks like a side-effect of the bundling process
- It is possible to implement the ```sqlite3_version``` method as a single line, on the same line as the ```gem 'sqlit3'``` declaration, but it would be a very long line.
- I did look at using GEM::Version to compare the different versions, so this code would not need to be modified with Rails 6-1-stable is released, but there are differing formats e.g. ```6-0-stable``` vs ```5.2.1``` used in travis.yml, so it adds further complication.

Please let me know if you would like me to make any changes.

### DIFF
--- a/ransack.gemspec
+++ b/ransack.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'rspec', '~> 3'
   s.add_development_dependency 'machinist', '~> 1.0.6'
   s.add_development_dependency 'faker', '~> 0.9.5'
-  s.add_development_dependency 'sqlite3', ENV['RAILS'] == '6-0-stable' ? '~> 1.4.1' : '~> 1.3.3'
+  s.add_development_dependency 'sqlite3', !ENV['RAILS'] || ENV['RAILS'] == '6-0-stable' ? '~> 1.4.1' : '~> 1.3.3'
   s.add_development_dependency 'pg', '~> 0.21'
   s.add_development_dependency 'mysql2', '0.3.20'
   s.add_development_dependency 'pry', '0.10'
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@ rvm:
   - 2.5
 
 env:
+  - DB=sqlite3
+  - DB=mysql
+  - DB=postgres
+
   - RAILS=6-0-stable DB=sqlite3
   - RAILS=6-0-stable DB=mysql
   - RAILS=6-0-stable DB=postgres
--- a/polyamorous/polyamorous.gemspec
+++ b/polyamorous/polyamorous.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'rspec', '~> 3'
   s.add_development_dependency 'machinist', '~> 1.0.6'
   s.add_development_dependency 'faker', '~> 1.6.5'
-  s.add_development_dependency 'sqlite3', '~> 1.3.3'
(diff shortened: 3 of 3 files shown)


ransack.gemspec
```diff
@@ -2,6 +2,11 @@
 $:.push File.expand_path("../lib", __FILE__)
 require "ransack/version"
 
+def sqlite3_version
+  return '~> 1.4.1' if ENV['RAILS'].nil? || ENV['RAILS'] == '6-0-stable'
```
gregmolnar: It would be neater to make this a one-liner:
`!ENV['RAILS'] || ENV['RAILS'] == '6-0-stable' ? '~> 1.4.1' :  '~> 1.3.3'`

obromios: I will do that. Should it also go into the polyamorous gemspec as well, given there are plans to package that as a gem separately?

gregmolnar: Yes, doing it for polyamorous would be also great. Could you please then just rebase it to 2 commits, one to fix the polyamorous gemspec and another to fix the ransack one? Thanks!

obromios: Will do, hopefully by next weekend.

obromios:
As you can see the CI tests are failing, yet the CI tests on my fork of this repository all passed for the same push. I have asked a[ question](https://stackoverflow.com/q/57237227/1299362) on SO, but no answer yet.  Are you seeing any CI test problems on this repository? If not, and if no one answers the SO question, then I suggest that I close this PR, create a new one, and push the changes to that new PR.

obromios:
As you can see the CI tests are failing, yet the CI tests on my fork of this repository all passed for the same push. I have asked a [question](https://stackoverflow.com/q/57237227/1299362) on SO. Have you seen any issue with CI test on this repository? If not, and if no one answers my question, I suggest I close this PR, open a new new PR, and push the changes to the new PR.

obromios:
There has been light thrown on why this difficulty has occurred, so I will close this PR and open a new one.

PR #1046, opened 2019-07-29
Adjust gemspecs for rails 6 default

### LINKED ISSUE #1042: Difficulty running tests on repository

obromios:
This is a pull request to fix #1042. This issue means that
the following commands

```
bundle install
bundle exec rake spec
raise an exception when run on a cloned master branch of the repository.
```

The problem was caused by a change to the default version of Rails. This change meant that the ransack gemspec incorrectly sets the version of sqlite3 when the RAILS env variable is undefined.

The fix was to alter the gemspec to set the correct version of sqlite3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@ rvm:
   - 2.5
 
 env:
+  - DB=sqlite3
+  - DB=mysql
+  - DB=postgres
+
   - RAILS=6-0-stable DB=sqlite3
   - RAILS=6-0-stable DB=mysql
   - RAILS=6-0-stable DB=postgres
--- a/polyamorous/polyamorous.gemspec
+++ b/polyamorous/polyamorous.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'rspec', '~> 3'
   s.add_development_dependency 'machinist', '~> 1.0.6'
   s.add_development_dependency 'faker', '~> 1.6.5'
-  s.add_development_dependency 'sqlite3', '~> 1.3.3'
+  s.add_development_dependency 'sqlite3', !ENV['RAILS'] || ENV['RAILS'] == '6-0-stable' ? '~> 1.4.1' : '~> 1.3.3'
 
   s.files         = `git ls-files`.split("\n")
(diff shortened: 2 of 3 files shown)


obromios:
This is a totally new fork, with the two commits pushed without any forcing, and I am still seeing this error as Travis CI attempts to build the system.
```
$ mysql -e 'create database ransack collate utf8_general_ci;'
ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)
The command "mysql -e 'create database ransack collate utf8_general_ci;'" failed and exited with 1 during .
```
This time I am seeing the same error on my fork and what is more, if I do a build on your master branch, I see the same problem.  Also note that I was able to do these builds on the earlier PR successfully, prior to the last step which was to rebase so there were only two commits.  This time there is no rebase, so I think this is a new problem related to the build script.

I think this it is most likely with the line:
```
mysql -e 'create database ransack collate utf8_general_ci
```
line which is a before_script instruction. Or it could be a Travis issue, there has been a [similar issue](https://github.com/travis-ci/travis-ci/issues/6842) before.  Do you think I should raise a travis issue? If so, can you please do a build on your master branch to see if it is happening to you as well?

gregmolnar:
This one if fixed on master now.

PR #1053, opened 2019-08-16
Update constants.rb

kobaltz:
Rails 6.0.0 has been released. Updating an existing 6.0.0.RC2 application had an issue with the constant.


gregmolnar:
This one is covered in https://github.com/activerecord-hackery/ransack/pull/1056

kobaltz:
Good catch! Thanks!

PR #1055, opened 2019-08-17
Fix CI setup

deivid-rodriguez:
I'm guessing because of infrastructural changes in TravisCI, `mysql` does not seem started by default, so the CI is failing in the setup step.

This PR makes sure mysql is started explicitly.

With this PR CI no longer fails in the `before_script` step, but instead shows a bunch of failures later on, that should get fixed by other PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,3 +40,6 @@ addons:
   code_climate:
     repo_token: 8b701c4364d51a0217105e08c06922d600cec3d9e60d546a89e3ddfe46e0664e
   postgresql: "9.6"
+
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
@gregmolnar I noticed that you've included the same change in the PR you just opened, so I guess this is not getting merged, right?

gregmolnar:
@deivid-rodriguez yes, I didn't see this PR before open that. Thanks for your work anyways!

deivid-rodriguez:
No problem!

PR #1058, opened 2019-08-19
Fix incompatibility with Dependabot

nguyenductung:
Fix #1059

PR #1060, opened 2019-08-19
Refactor gemspec

deivid-rodriguez:
Hei! I initially prepared this patch to fix not being able to point to ransack's master branch from my Gemfile unless I define an `ENV["RAILS"]` environment variable.

That seems fixed by #1058, but in any case, this implements a set of better practices when managing your gemspec.

In general, development dependencies fit better in your Gemfile, because they are only considered for development, an never affect end users like it happened here. Also, gemspecs don't have the feature of "conditional dependencies", which was probably the intention for the `sqlite3` dependency here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: ruby
 sudo: false
 
 rvm:
-  - 2.6.0
+  - 2.6.3
 
 services:
   - mysql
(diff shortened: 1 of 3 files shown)


gregmolnar:
Thank you!

deivid-rodriguez:
Thanks!

PR #1070, opened 2019-10-13
Drop old rubies support

deivid-rodriguez:
I want to create a PR to fix some issues with (unreleased) ruby 2.7, but I want to use String#+@, which is only available since ruby 2.3.

I can create a different patch (backwards compatible with all currently supported rubies) using `String#dup`, but since ruby 1.9, 2.0, 2.1 and 2.2 reached their end of life a while ago, I figure we could drop support for those.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Drop support for rubies under 2.3.
+  PR [#1070](https://github.com/activerecord-hackery/ransack/pull/1070)
+
 ## 2.3.0 - 2019-08-18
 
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
Also, note that although installation was permitted on old rubies, README currently notes that the gem only works on ruby 2.2 or higher. So this PR might only be effectively dropping support for ruby 2.2.

deivid-rodriguez:
Ping! Improved and rebased.

deivid-rodriguez:
Rebased this one and all of my other PRs now that CI is green :+1:.

n-rodriguez:
Hi there! Any news?

PR #1072, opened 2019-10-14
Fix running tests against a Rails clone

deivid-rodriguez:
There were two problems:

* An absolute path is not a something `Gem::VERSION` understands, so in
the "path case", read the `RAILS_VERSION` file from the Rails clone and
use that.

* In the "path case" there were some missing dependencies. Add them to
the Gemfile.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,8 +5,17 @@ gem 'rake'
 
 rails = ENV['RAILS'] || '6-0-stable'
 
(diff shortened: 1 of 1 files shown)


seanfcarroll:
This has been merged, thank you @deivid-rodriguez

deivid-rodriguez:
Thanks to you! :purple_heart:

PR #1074, opened 2019-10-16
Make Actionpack optional dependency

vimutter:
Hi, in some cases we don't need view and controller layer of rails but still want to use ransack with AR.
This PR makes actionpack an optional dependency. So if we don't need it it won't be required and added to Gemfile.lock


ransack.gemspec
```diff
@@ -14,12 +14,13 @@ Gem::Specification.new do |s|
...
   s.license     = 'MIT'
 
-  s.add_dependency 'actionpack', '>= 5.0'
+  
   s.add_dependency 'activerecord', '>= 5.0'
   s.add_dependency 'activesupport', '>= 5.0'
   s.add_dependency 'i18n'
   s.add_dependency 'polyamorous', Ransack::VERSION.to_s
-
+  s.add_development_dependency 'actionpack', '>= 5.0'
```
deivid-rodriguez: Development dependencies were moved to the `Gemfile` in https://github.com/activerecord-hackery/ransack/commit/8c92a3b1f6f2360599732f8d1ff4ae95f515eb86, and `actionpack` is already present in the current Gemfile, so you can remove this addition.

ransack.gemspec
```diff
@@ -13,13 +13,12 @@ Gem::Specification.new do |s|
...
   s.license     = 'MIT'
-
-  s.add_dependency 'actionpack', '>= 5.0'
+  
   s.add_dependency 'activerecord', '>= 5.0'
   s.add_dependency 'activesupport', '>= 5.0'
   s.add_dependency 'i18n'
   s.add_dependency 'polyamorous', Ransack::VERSION.to_s
-
+
```
deivid-rodriguez: I would cleanup this trailing whitespace you're introducing here.

seanfcarroll:
@vimutter thank you for this contribution. Any chance of a rebase and little cleanup?

seanfcarroll:
Thanks very much @vimutter and @deivid-rodriguez !

PR #1076, opened 2019-10-16
Allow configuring test DB connection

deivid-rodriguez:
Provide a little more flexibility when configuring the test DB, so developers with custom configuration don't need to edit the source files.


seanfcarroll:
I love this feature @deivid-rodriguez, but could I ask you to add a test please?

deivid-rodriguez:
mmmm, but this is just a testing utility, it's not used by end users.

I only added this as a convenience for myself (and maybe others?) when running ransack tests locally, because I have my local postgres configuration set to "md5" instead of "trust".

seanfcarroll:
OK fair enough

Issue #1039, opened 2019-07-03
undefined method `polymorphic?' for ActiveRecord::Reflection::PolymorphicReflection


PhilCoggins:
I am using Rails 5.2.3, and Ransack latest master(e17f1c60a0654517e75d6502aa57b50583b27e6a) and have found a potential bug.

It looks like this issue surfaces when trying to filter by a `has_many through` relationship where the `source` is a polymorphic `belongs_to`.

I am getting the below error:
```
undefined method `polymorphic?' for ActiveRecord::Reflection::PolymorphicReflection
# ./polyamorous/lib/polyamorous/activerecord_5.2.0_ruby_2/reflection.rb:4:in `build_join_constraint'
     # /Users/philcoggins/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/bundler/gems/rails-5dbae010a9d4/activerecord/lib/active_record/associations/join_dependency/join_association.rb:33:in `block in join_constraints'
     # /Users/philcoggins/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/bundler/gems/rails-5dbae010a9d4/activerecord/lib/active_record/associations/join_dependency/join_association.rb:29:in `reverse_each'
     # /Users/philcoggins/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/bundler/gems/rails-5dbae010a9d4/activerecord/lib/active_record/associations/join_dependency/join_association.rb:29:in `with_index'
     # /Users/philcoggins/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/bundler/gems/rails-5dbae010a9d4/activerecord/lib/active_record/associations/join_dependency/join_association.rb:29:in `join_constraints'
     # ./polyamorous/lib/polyamorous/activerecord_5.2.1_ruby_2/join_dependency.rb:53:in `make_constraints'
     # /Users/philcoggins/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/bundler/gems/rails-5dbae010a9d4/activerecord/lib/active_record/associations/join_dependency.rb:183:in `block in walk'
     # /Users/philcoggins/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/bundler/gems/rails-5dbae010a9d4/activerecord/lib/active_record/associations/join_dependency.rb:183:in `each'
[...]

PR #1077, opened 2019-10-16
Fix issue evaluating polymorphic joins

deivid-rodriguez:
Fixes https://github.com/activerecord-hackery/ransack/issues/1039, inspired on @PhilCoggins's investigation in https://github.com/activerecord-hackery/ransack/issues/1039#issue-463944173, but being initially more aggressive.


PhilCoggins:
@deivid-rodriguez Thank you for taking a stab at this. Luckily I wrote tests wrapping our use cases here, so I was able to pull your branch and run a few of them.

I also initially thought that removing Polyamorous for 5.2 would "just work", but somewhere along the way the condition to constrain `notable_type` gets dropped. Without that condition, we risk joining in polymorphic records of an unintended type (this happened to us in production).

I can get a failing test by adding this line to https://github.com/deivid-rodriguez/ransack/blob/fix_rails_52_polymorphic_joins/spec/ransack/search_spec.rb#L288

```
# Expect the query to constrain notable_type
expect(s.to_sql).to match /"notes"."notable_type" = 'Person'/
```

Failure:
```
       expected "SELECT \"notes\".* FROM \"notes\" LEFT OUTER JOIN \"people\" ON \"people\".\"id\" = \"notes\".\"notable_id\" WHERE \"people\".\"name\" = 'Ernie'" to match /"notes"."notable_type" = 'Person'/
       Diff:
       @@ -1,2 +1,2 @@
       -/"notes"."notable_type" = 'Person'/
       +"SELECT \"notes\".* FROM \"notes\" LEFT OUTER JOIN \"people\" ON \"people\".\"id\" = \"notes\".\"notable_id\" WHERE \"people\".\"name\" = 'Ernie'"
```

My monkey-patch in the original issue accounted for this in the first `if` condition:
```
      if respond_to?(:polymorphic?) && polymorphic?
        super(table, foreign_table)
        .and(foreign_table[foreign_type].eq(klass.name))
```

With this patch, the query in the failing test should look like:
```
SELECT notes
FROM notes
LEFT OUTER JOIN people ON people.id = notes.notable_id
AND notes.notable_type = 'People'
```

Emphasis on the `AND` condition in the join. Hopefully this makes sense, LMK if this is unclear.

deivid-rodriguez:
Thank you @PhilCoggins!

So, the code the posted in the issue should be correct, and we also have a regression test for it, right?

Can you create a PR and I'll close this one, I don't want to steal attribution, you did all the work here :)

PhilCoggins:
@deivid-rodriguez The code posted in the issue works with ActiveRecord v5.2.3, but fails on 5-2 stable, see https://github.com/rails/rails/commit/54de9b1e4f5823bf22001be60efcb996ce6d260b.
That commit unfortunately removes the `build_join_constraint` method being overridden in my original patch.

I've spent nearly a full day going back and forth between ActiveRecord and Polyamorous and I'm afraid I just can't figure it out 😞. I appreciate you trying to attribute the fix to me, but I really just want someone who already understands the internals of this library to fix it.

deivid-rodriguez:
Ok! I'll try to dig in more and see if we can fix this.

deivid-rodriguez:
I plan to (at least try to) figure this out soon :+1:

seanfcarroll:
@deivid-rodriguez based on your latest comment this is not ready to merge? I'm wondering why the CI is passing.

deivid-rodriguez:
No, it's not.

If I recall correctly, CI is passing because the assertion in my spec is too soft (needs to be strenghten by the assertion proposed by @PhilCoggins), and because tests against 5-2-stable and 6-0-stable are currently in the "allowed_failures" section, so failures there are "hidden".

oboxodo:
Hi guys. I'm sorry to be a PITA. We're trying to upgrade our project from rails 5.2.x to 6.0 and this requires us to upgrade ransack from 2.1.1 to 2.3.0. While doing that, some of our tests are failing, apparently because of the bug this patch might solve.

Has anyone been able to use ransack 2.3.0 with rails 5.2.x and rails 6.0.x successfully without this patch?

seanfcarroll:
@oboxodo please see this comment https://github.com/activerecord-hackery/ransack/issues/1081#issuecomment-573236434

deivid-rodriguez:
Thanks for fixing this @PhilCoggins!!

kul-p:
I have this "ArgumentError: wrong number of arguments (given 3, expected 2)"
from 
from /Users/vkul9pr/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/polyamorous-2.3.2/lib/polyamorous/activerecord_6.0_ruby_2/join_dependency.rb:32:in `join_constraints'

Don't sure it's related with one you all discuss 
I am using rails 6.1 (not have this issue in rails 6.0)
Hope any one can fix this/ Thanks

PR #1089, opened 2019-11-25
Improves ransackable_scopes_skip_sanitize_args doc

eikes:
The readme should make clear, that the ransackable_scopes_skip_sanitize_args method is a model class method, not an instance method.


seanfcarroll:
Thanks for this contribution @eikes

PR #1101, opened 2020-01-13
specify actual version of polyamorous, so we can release that separately

gregmolnar:


seanfcarroll:
@gregmolnar @tvdeyen 

Do we want to now bump both Ransack and the Polyamorous inside Ransack to 2.3.3 and push both to Rubygems?

https://github.com/activerecord-hackery/ransack/pull/1104

gregmolnar:
We only need to release ransack now, it will depend on polyamorous `2.3.2` which is already released.

PR #1104, opened 2020-01-13
Only include necessary files in gem package

### LINKED ISSUE #1102: Packaged gem includes polyamorous code twice
The `ransack` gem package includes the code of the `polyamorous` code twice because of this line 

https://github.com/activerecord-hackery/ransack/blob/master/ransack.gemspec#L22

Since polyamorous is shipped as its own gem we now have the polyamorous code twice.

We should exclude the files from the ransack gem package.

tvdeyen:
This gem also includes the polyamorous gem (for now). We must not include its code inside the ransacks gem package.

Closes #1102

### DIFF
--- a/ransack.gemspec
+++ b/ransack.gemspec
@@ -19,14 +19,8 @@ Gem::Specification.new do |s|
   s.add_dependency 'i18n'
   s.add_dependency 'polyamorous', Ransack::VERSION.to_s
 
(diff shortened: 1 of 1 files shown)


seanfcarroll:
Thank you @tvdeyen !

tvdeyen:
@seanfcarroll any change to release this on rubygems? The current releases have the duplicated code in them.

Issue #1099, opened 2020-01-12
Convert Polyamorous into a module


seanfcarroll:
Previously Polyamorous was a [separate gem](https://github.com/activerecord-hackery/polyamorous). It has been moved into the Ransack repo, which is a great step forward. 

However, it is still a gem **_inside_** the Ransack repo. What needs to be done is to convert this code purely into a module. A later refactoring can more closely merge the code, but for now the work needed is:

- Remove the Polyamorous gem-specific code
   - https://github.com/activerecord-hackery/ransack/blob/master/polyamorous/polyamorous.gemspec
  - https://github.com/activerecord-hackery/ransack/blob/master/polyamorous/lib/polyamorous/version.rb
- Add any dependencies needed into the Ransack gemspec
- Move everything under the main `/lib/polyamorous` folder
- Remove the Polyamorous dependency from Ransack
  - https://github.com/activerecord-hackery/ransack/blob/master/ransack.gemspec#L20
- Add the [Polyamorous tests](https://github.com/activerecord-hackery/polyamorous/tree/master/spec) into the Ransack repo and get them to pass

seanfcarroll:
I'm thinking the Polyamorous gem should be removed from Rubygems also. Any thoughts on this?

tvdeyen:
> I'm thinking the Polyamorous gem should be removed from Rubygems also. Any thoughts on this?

I think we should release polyamorous 2.3.1 and 2.3.2 so that ransack 2.3.1 and 2.3.2 can actually be installed via bundler (#1098)

Then release a ransack 3.0 with included polyamorous (once this issue has been resolved) and leave the old polyamorous gems on rubygems for old projects and do not release a 3.0

That way we do not break peoples CIs and local development

tvdeyen:
And probably include some deprecation warnings in the 2.3.2 releases of polyamorous telling the people about the removal and inclusion in ransack 3.0

tvdeyen:
@seanfcarroll I am a little confused about this issue, actually. In #1002 you decided to make polyamorous a seperate gem. There was some discussion around to use polyamorous with other gems as well (https://github.com/rzane/baby_squeel/issues/97).

Is this issue about reverting #1002? I am willing to work on this, but I am not sure what is the goal here? Maybe @gregmolnar can shed some light?

gregmolnar:
@tvdeyen see my reply here: https://github.com/activerecord-hackery/ransack/issues/1098#issuecomment-573617666

[...]

Issue #1081, opened 2019-10-21
CI against Rails 6-0-stable broken


deivid-rodriguez:
CI is broken against 6-0-stable since https://github.com/rails/rails/commit/dd46d18e2108f2e953e9811c6eabf5b18fe77ab1.

Needs to be investigated and fixed.

seanfcarroll:
I'll try to look at this soon.

macfanatic:
Anything we can do to help mitigate or debug this issue @seanfcarroll?

We have confirmed the issue remains with rails releases `6.0.1.rc1` and `6.0.1.rc2`.

seanfcarroll:
Thanks for the ping @macfanatic 

Sorry I haven't had time to work on it. I will be able to look at it during the Christmas week. If you are interested in making a PR with any code investigations or failing tests for this specific problem, that could be helpful.

All the best 👍

macfanatic:
Here is a rails bug script showing that with rails `6.0.0` the test passes, but when changing to activerecord `6.0.1` the test fails with an uncaught invalid SQL exception.

```ruby
# frozen_string_literal: true
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "activerecord", "6.0.0", require: "active_record"
  gem "ransack", "2.3.0"
  gem "sqlite3"
end

require "minitest/autorun"
require "logger"

ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  create_table :organizations, force: true do |t|
    t.string :name, null: false
    t.string :type, null: false
    t.timestamps
  end

  create_table :products, force: true do |t|
    t.string :name, null: false
    t.references :seller, foreign_key: { to_table: :organizations }, null: false
    t.references :manufacturer, foreign_key: { to_table: :organizations }, null: false
    t.timestamps
  end
end

class Organization < ActiveRecord::Base
  validates :name, presence: true
end

class Seller < Organization
  has_many :products
end

class Manufacturer < Organization
  has_many :products
end

class Product < ActiveRecord::Base
  belongs_to :seller
  belongs_to :manufacturer

  validates :name, presence: true
end

class BugTest < Minitest::Test
  def test_generated_sql_join

    manufacturer = Manufacturer.create!(name: 'ABC Toys, LLC')
    seller = Seller.create!(name: 'Amazon, Inc')
[...]

PR #1105, opened 2020-01-21
Moved polyamorous from gem to a module

shadabmalik1310:
Closes https://github.com/activerecord-hackery/ransack/issues/1099


seanfcarroll:
Thanks @shadabmalik1310 I will review this later today. 

Just some initial notes before the review:

- we need to also include the [tests from Polyamorous](https://github.com/activerecord-hackery/polyamorous/tree/master/spec) that were previously missed. https://github.com/activerecord-hackery/ransack/issues/1099

- all the tests will need to pass. This might be dependent on https://github.com/activerecord-hackery/ransack/issues/1081

Many thanks!

ransack.gemspec
```diff
@@ -16,11 +16,11 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'activerecord', '>= 5.2.1'
   s.add_dependency 'activesupport', '>= 5.2.1'
+  s.add_dependency 'machinist', '~> 1.0.6'
```
deivid-rodriguez: Why is this runtime dependency being added? :thinking:

spec/ransack/adapters/active_record/context_spec.rb
```diff
@@ -91,22 +91,22 @@ module ActiveRecord
...
-            end
-
-            it 'can be rejoined to execute a valid query' do
-              parents, children = shared_context.join_associations
-
-              expect { Person.joins(parents).joins(children).to_a }
-              .to_not raise_error
-            end
-          end
+          # describe '#join_associations', if: AR_version <= '4.0' do
```
seanfcarroll: As we are not supporting AR <= 4 this test can be removed.

spec/ransack/search_spec.rb
```diff
@@ -234,12 +234,21 @@ module Ransack
...
           parent_articles_title_eq: 'parents_article_title_query'
         }).result
+
         real_query = remove_quotes_and_backticks(s.to_sql)
 
-        expect(real_query)
-          .to match(%r{LEFT OUTER JOIN articles ON (\('default_scope' = 'default_scope'\) AND )?articles.person_id = people.id})
-        expect(real_query)
-          .to match(%r{LEFT OUTER JOIN articles articles_people ON (\('default_scope' = 'default_scope'\) AND )?articles_people.person_id = parents_people.id})
+        if ::Gem::Version.new(::ActiveRecord::VERSION::STRING) > ::Gem::Version.new(Constants::RAILS_6_0)
```
seanfcarroll: Can we also add an a test where the data for this condition is created, and we check that the output is the same across each version of AR.

ransack.gemspec
seanfcarroll: Thanks for the review @deivid-rodriguez !

Has now been removed.

spec/spec_helper.rb
```diff
@@ -1,4 +1,5 @@
 require 'machinist/active_record'
+require 'polyamorous/polyamorous.rb'
```
seanfcarroll: We need the specs from the Polyamorous gem added and passing.

spec/ransack/adapters/active_record/context_spec.rb
shadabmalik1310: ok

spec/ransack/search_spec.rb
shadabmalik1310: I have tried but i couldn't find a way to change the dynamically change the version of AR. But i have tested with different versions without spec, it's fine.

spec/spec_helper.rb
shadabmalik1310: ok

shadabmalik1310: Hi @seanfcarroll - All files are covered with the specs and coverage is 100%, let me know the file and line which is not covered with the spec.

Thanks

spec/ransack/search_spec.rb
seanfcarroll: Thanks @shadabmalik1310 the way to do this would to:

- set up some data 
- check for an expected output of a Ransack query as data (not the SQL generated)
- run it in CI for each different version of Rails

spec/spec_helper.rb
seanfcarroll: Could you please add in the code missing from here https://github.com/activerecord-hackery/polyamorous/blob/master/spec/polyamorous/join_association_spec.rb#L20-L37

Also, please move the Polyamorous tests under their own folder (at the same level as Ransack).

Thank you. :-)

seanfcarroll:
Thanks for the work @shadabmalik1310 

If the review points are all now resolved, would it be possible to make a note and then resolve the conversation?

Let's get this merged!

seanfcarroll:
bump @shadabmalik1310

seanfcarroll:
bump @shadabmalik1310

seanfcarroll:
Closed as superceeded by https://github.com/activerecord-hackery/ransack/pull/1113

PR #1121, opened 2020-05-08
Fix Ruby 2.7 deprecation warnings

terracatta:
In Ruby 2.7 passing the last argument as an options hash if the receiver uses keyword arguments is no longer allowed and throws a warning.

This PR applies the recommended fix of adding `**` to covert the options hash into the correct keyword arg format.

There may be more deprecation warnings to fix in other parts of the codebase but these were the ones that jumped out to me right away.


foton:
I would like to see this merged. :+1:

seanfcarroll:
Thank you for your contribution @terracatta

PR #1132, opened 2020-06-08
Add ransack! method which raises error if passed unknown condition

alipman88:
**Use case:** We use Ransack for both public-facing search pages and internal purposes such as targeting bulk email deliveries. In the context of public-facing search, we'd like to be lenient with unknown/invalid search params. However, in the context of targeting bulk emails, this seems dangerous - a broken query could potentially go out to many more recipients than intended.

Thus, we'd like to be able to trigger the `ignore_unknown_conditions` config option on a case-by-case basis. (Previously requested in https://github.com/activerecord-hackery/ransack/issues/535)

I've drafted what feels like a clean implementation, although here are some other options I considered:

```ruby
# add Ransack::Search#validate!
Person.ransack(unknown_attr_eq: "Ernie").validate!.result

# pass ignore_unknown_conditions directly as param
Person.ransack(unknown_attr_eq: "Ernie", ignore_unknown_conditions: false)
```

If folks like this idea but have a different preference for implementing it, let me know and I'll adjust this PR accordingly!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+* Add `ActiveRecord::Base.ransack!` which raises error if passed unknown condition
+
+  *Aaron Lipman*
+
 ## 2.4.0 - 2020-11-27
 
 *
--- a/README.md
+++ b/README.md
@@ -670,6 +670,43 @@ Trying it out in `rails console`:
 
 That's it! Now you know how to whitelist/blacklist various elements in Ransack.
 
+### Handling unknown predicates or attributes
+
+By default, Ransack will ignore any unknown predicates or attributes:
+
+```ruby
+Article.ransack(unknown_attr_eq: 'Ernie').result.to_sql
+=> SELECT "articles".* FROM "articles"
+```
+
+Ransack may be configured to raise an error if passed an unknown predicate or
+attributes, by setting the `ignore_unknown_conditions` option to `false` in your
+Ransack initializer file at `config/initializers/ransack.rb`:
+
+```ruby
+Ransack.configure do |c|
+  # Raise errors if a query contains an unknown predicate or attribute.
+  # Default is true (do not raise error on unknown conditions).
+  c.ignore_unknown_conditions = false
+end
+```
+
+```ruby
+Article.ransack(unknown_attr_eq: 'Ernie')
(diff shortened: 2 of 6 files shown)


seanfcarroll:
@alipman88 thank you for this PR. 🎸 

Can I suggest you rebase as there have been a lot of fixes to Ransack, so this might pass the CI now.

It is a nice feature, and I'm wondering if as part of this PR the `ignore_unknown_conditions` configuration option could also be documented in the README ? If there are other undocumented configuration options, perhaps also those?

alipman88:
Getting back around to this - thanks for your patience, @seanfcarroll.

> I'm wondering if as part of this PR the `ignore_unknown_conditions` configuration option could also be documented in the README ?

Good catch. I've updated the readme to include include both the `ignore_unknown_conditions` and the new `ransack!` method under a "Handling unknown predicates or attributes" heading, which I've also moved a little further down (under the whitelisting/blacklisting section)".

> If there are other undocumented configuration options, perhaps also those?

I took a glance at the options in `lib/ransack/configuration.rb`. The remaining config options (`search_key`, `ignore_unknown_conditions`, `custom_arrows`, `sanitize_custom_scope_booleans` and `hide_sort_order_indicators`) are covered in the readme.

The slight exception is the `search_key` option, which is linked to from the the "If you're coming from MetaSearch" section of the readme. I'd say that section of the readme can be moved to the end of the readme now that MetaSearch is eight years old. I've added a second commit moving that section to the bottom, and documenting the `search_key` option elsewhere.

> this might pass the CI now.

The mysql tests are failing. Don't think that's related, but taking a quick look.

alipman88:
~The mysql tests are failing. Don't think that's related, but taking a quick look.~ nvm - I have no idea how, but I had commented out the `mysql2` line from the Gemfile. Fixed!

seanfcarroll:
This looks pretty useful, what do you think @deivid-rodriguez ?

deivid-rodriguez:
I'm sorry @seanfcarroll, I don't have time at the moment to evaluate this. If it looks good to you, please move it forward, otherwise I can give it a look when I find time :+1:.

seanfcarroll:
> nvm - I have no idea how, but I had commented out the mysql2 line from the Gemfile. Fixed!

@alipman88 😄 that will do it!

PR #1145, opened 2020-08-26
Make gem compatible with old polyamorous require

rtweeks:
By providing `polyamorous.rb` in the `lib` folder, projects that
previously depended directly on `polyamorous` can just switch the
dependency to `ransack` without changing their calls to `require`.


seanfcarroll:
Looks reasonable @rtweeks can you please rebase?

rtweeks:
Rebased on d508857.

PR #975, opened 2018-10-27
deprecate #search

gregmolnar:
Closes #949

PR #1147, opened 2020-09-08
Remove deprecated `#search` method

y-yagi:
`#search` is deprecated in v2.1.1 and it scheduled to remove the method in v2.3.
https://github.com/activerecord-hackery/ransack/pull/975

But the method still exists in v2.3.2. So let's remove the method.

If we removed the method, users can remove the patch for `search` (`ActiveRecord::Base.class_eval('remove_method :search'`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,32 +352,6 @@ construct much more complex search forms, such as the one on the
 [demo app](http://ransack-demo.herokuapp.com/users/advanced_search)
 (source code [here](https://github.com/activerecord-hackery/ransack_demo)).
 
-### Ransack #search method
-
-Ransack will try to make the class method `#search` available in your
-models, but if `#search` has already been defined elsewhere, you can always use
-the default `#ransack` class method. So the following are equivalent:
-
-```ruby
-Article.ransack(params[:q])
-Article.search(params[:q])
-```
-
-Users have reported issues of `#search` name conflicts with other gems, so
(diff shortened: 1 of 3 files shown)


y-yagi:
Hmm. Tests against 6-0-stable don't pass in master branch also.

yahonda:
Test failure against 6-0-stable is triggered by rails/rails#39268 .

yahonda:
I've just found this failure but have no idea how to fix it yet. Also, this method should be Rails non-public API which can be changed without warnings.

yahonda:
FYI https://github.com/activerecord-hackery/ransack/pull/1171 addresses the CI failure.

seanfcarroll:
@y-yagi please rebase and rerun the CI

y-yagi:
@seanfcarroll Done!

seanfcarroll:
Thanks @y-yagi as this is removing deprecated functions, it should be on a major version bump, ie: 3.0.0

Will do it when that version is released.

seanfcarroll:
@y-yagi FYI we can probably do a major version bump soon, so thank you for being patient.

deivid-rodriguez:
@y-yagi Can you rebase this? I think we can proceed with Ransack 3.0.0 now and include this PR.

y-yagi:
@deivid-rodriguez Done!

PR #1149, opened 2020-09-15
Use unfrozen version of symbol to string

fauno:
This change makes Ransack compatible with shopify/symbol-fstring since `Symbol#to_s` will return a frozen string literal.


seanfcarroll:
@fauno please rebase and try the CI again.

Would it be possible to add some more description to the PR? 

Are the any additional tests that could be added here?

fauno:
Thanks! Sorry I forgot to reply here!

seanfcarroll:
No problem @fauno thanks for your contribution!

PR #1155, opened 2020-10-23
Typo in docs

brett-anderson:
I think this is a typo, should be `ILIKE`, since it's case insensitive.


deivid-rodriguez (approved):
Thanks!!

PR #1157, opened 2020-11-01
Fix typo - change LIKE to ILIKE for i_cont in README

coreyaus:
The i_cont predicate runs a case insensitive search: https://github.com/activerecord-hackery/ransack/blob/master/lib/ransack/adapters/active_record/ransack/constants.rb#L19


coreyaus:
Oops - closing as duplicate of https://github.com/activerecord-hackery/ransack/pull/1155 😊

PR #1169, opened 2020-11-20
Test against Rails 6.0.3

deivid-rodriguez:


deivid-rodriguez:
Rails 6.0.3 seems good!

seanfcarroll:
If there is a minor Rails update, ie: `6.0.3`, should we drop support for `6.0.0`? 🤔 

The CI would then be green ✅

deivid-rodriguez:
https://github.com/activerecord-hackery/ransack/pull/1171 should get CI back on track.

PR #1178, opened 2020-12-01
Links to Tidelift subscription

deivid-rodriguez:
I'm completing a few tasks that I need to setup Tidelift, like linking to their subscription in a FUNDING.yml file or linking to their coordinated disclosure plan for security issues.


seanfcarroll:
This looks good to me @deivid-rodriguez, merging.

PR #1176, opened 2020-11-28
Add spec described in #1153.

MrJoy:
This adds a spec that should demonstrate the issue described in #1153.

PR #1180, opened 2020-12-01
Enable GitHub Actions

yahonda:
This pull request enables GitHub Actions to run unit tests.

- Use ubuntu-20.04
- Use Ruby 2.7
- CI against SQLite3, MySQL and PostgreSQL databases
- CI against Rails v6.1.0.rc1, v6.0.3, 6-0-stable, 5-2-stable and v5.2.4
- Add MYSQL_USERNAME and MYSQL_PASSWORD environment variables

Refer https://github.com/activerecord-hackery/ransack/pull/1176#issuecomment-735644447 for the background of this pull request.


deivid-rodriguez (changes requested):
Thanks @yahonda, this is awesome! :tada: 

Could you also replace the Travis badge in the README with Github Actions, and replace any other mentions to TravisCI in the repo?

Also, it's great to verify that ransack works fine with ruby 2.7, but I guess we should keep testing against 2.6.6 just in case. Could add a matrix.ruby dimension and test both?

Finally, and as a little nitpick, could you use an exact version, like 2.7.2. It's a personal preference, and it's what was previously used in the TravisCI configuration, so we should keep it that way.

deivid-rodriguez:
And, and we should also completely delete the previous travis configuration :)

yahonda:
Thanks for the review. Sure let me update this pull request as suggested.

README.md
```diff
@@ -4,7 +4,7 @@
 
 Please see the [Maintainer wanted issue](https://github.com/activerecord-hackery/ransack/issues/1159) if you are interested.
 
-[![Build Status](https://travis-ci.org/activerecord-hackery/ransack.svg)](https://travis-ci.org/activerecord-hackery/ransack)
+[![Build Status](https://github.com/activerecord-hackery/ransack/workflows/test/badge.svg)](https://github.com/activerecord-hackery/ransack/actions)
```
deivid-rodriguez, line 7: I think it's `https://github.com/activerecord-hackery/ransack/workflows/Test/badge.svg`. With lowercase I get "Not found".

deivid-rodriguez: Either that or change the workflow name to "test".

yahonda: Right. I prefer to use the lowercase "test".

deivid-rodriguez (approved):

deivid-rodriguez:
Merging to get a CI back on track!

deivid-rodriguez:
Thanks so much @yahonda!

PR #1189, opened 2020-12-25
Drop Ruby 2.5 or older versions of Ruby

yahonda:
README already mentions that `Ransack is supported for Rails 6.1, 6.0,
5.2 on Ruby 2.6.6 and later.` by https://github.com/activerecord-hackery/ransack/commit/d50885772cfa9c2534c06f7edb0eb4141f2e1a4a

I'd like to get some feed back from maintainers since #1070 just dropped Ruby 2.2 or older versions of Ruby.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
   *Aaron Lipman*
 
+* Drop support for rubies under 2.5. PR #1189
+
 ## 2.4.0 - 2020-11-27
 
 *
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
I'm happy with it, thanks!

I would add a changelog entry for it though.

yahonda:
Thanks for the review. Added a changelog entry.

deivid-rodriguez:
Thank you!

PR #1190, opened 2020-12-25
CI against Ruby 3.0.0

yahonda:
This pull request runs CI against Ruby 3.0.0.
Rails 6.0.3 or higher is required to support Ruby 3 keyword arguments behavior then excluded the matrix between Ruby 3.0.0 and Rails 5.2.x.

* Rails 6.0.3 has been released!
https://weblog.rubyonrails.org/2020/5/6/Rails-6-0-3-has-been-released/
> This version fixes warnings when used with Ruby 2.7.

* Ruby 3.0.0 Released
https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/
> Keyword arguments are separated from other arguments.

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: 2.7.2
+          ruby-version: 3.0.0
       - name: Install gems
         run: bundle install --jobs 4 --retry 3
       - name: Run RuboCop
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,8 +17,14 @@ jobs:
           - 5-2-stable
           - v5.2.4
         ruby:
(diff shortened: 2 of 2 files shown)


deivid-rodriguez (approved):
Looks good, thanks!

Should we consider dropping Rails 5.2 support?

yahonda:
Thanks for the review. In my opinion, Dropping Rails 5.2 support is still needed now.

deivid-rodriguez:
@yahonda Do you mean that "Rails 5.2 is still needed" or that "dropping Rails 5.2 support is still needed"? Couldn't understand if you're in favor or against the change :sweat_smile:.

yahonda:
Thanks for the confirmation. I meant to say "Rails 5.2 is still needed" .

deivid-rodriguez:
Ok, I don't have a special interest in dropping 5.2 support right now so let's leave it for now.

deivid-rodriguez:
Thanks for this!

PR #1192, opened 2020-12-30
Allow Ransack to be tested with Rails main branch

yahonda:
This pull request allows Ransack to be tested with Rails master branch aka 6.2.0.alpha.

Let me explain some background for this change.

Ransack (polyamorous) depends on some Rails internal APIs and Rails internal APIs change without deprecation warnings. In the long term, Ransack (polyamorous) may not depend on these internal API, but as of now, what we can do is run CI against Rails master branch so that we will able to know which commits 'break' Ransack behavior.

### DIFF
--- a/.github/workflows/cronjob.yml
+++ b/.github/workflows/cronjob.yml
@@ -16,7 +16,7 @@ jobs:
           - 2.6.6
     env:
       DB: sqlite3
-      RAILS: master
+      RAILS: main
     steps:
       - uses: actions/checkout@v2
       - name: Set up Ruby
@@ -39,7 +39,7 @@ jobs:
           - 2.6.6
     env:
       DB: mysql
-      RAILS: master
+      RAILS: main
       MYSQL_USERNAME: root
       MYSQL_PASSWORD: root
     steps:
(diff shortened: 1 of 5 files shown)


yahonda:
Let me take a look at `[!] There was an error parsing `Gemfile`: Malformed version number string master. Bundler cannot continue.` error.

yahonda:
`[!] There was an error parsing Gemfile: Malformed version number string master. Bundler cannot continue. error.` has been addressed via https://github.com/activerecord-hackery/ransack/pull/1192/commits/0d60b9a5da7bef4ad5e52a2f97859ec829d8dc45

deivid-rodriguez (commented):
I'm cool with this and I agree with you that it'd be nice to discover breakage as soon as it happens :+1:.

What I dislike is the fact of checking this for every PR, because I like PRs to be green for merging them, and doing this makes a PR potentially red at any time, without it being caused by the changes in a PR. My personal practice is that a PR should be green if and only if the changes in the PR are good.

What I like to do is using a daily scheduled worklfow for this. Currently github doesn't support builtin notifications for this, so if you want notifications you need to do a bit more work, but just having it run daily and being able to check the status in the actions tab helps I believe.

yahonda:
Thanks for the review and I understand what you do not like. Let me create another workflow called "cronjob" or something like that which executes daily cron job against Rails master branch.

seanfcarroll:
Are you happy for this to be merged now @deivid-rodriguez ?

deivid-rodriguez (approved):
Yes!

PR #1193, opened 2020-12-31
Bump RuboCop

yahonda:
This pull request bumps RuboCop version by removing version limitation because Ransack supports Ruby 2.6 or higher.

Also enabled `Style/RedundantFileExtensionInRequire` cop which is available since RuboCop 0.88.0.


deivid-rodriguez (approved):
Thanks!

PR #1194, opened 2021-01-03
Fix CHANGELOG [ci skip]

y-yagi:
* Add section for `2.4.1` release
* Remove empty entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,14 @@
 # Change Log
 
+* Drop support for rubies under 2.5. PR #1189
+
+## 2.4.1 - 2020-12-21
+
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Thanks @y-yagi!

@yahonda Do you know what changed so that CI is now red on ruby 3.0?

yahonda:
These failures are not relevant to this pull request itself.

It looks like Faker gem v1 does not support Ruby 3 keyword arguments. I have opened pull request #1195 to bump Faker gem version to 2.

seanfcarroll:
Thank for looking after this @y-yagi and @yahonda 🚀

PR #1184, opened 2020-12-09
Add ability to config PostgreSQL ORDER BY ... NULLS FIRST or NULLS LAST

itsalongstory:
Close #1125

Related to https://github.com/activerecord-hackery/ransack/issues/443#issuecomment-551390658

https://www.postgresql.org/docs/13/queries-order.html

PR #1196, opened 2021-01-06
Ensure relation.order method receives Arel to silence deprecation warning

Polidoro:
Ransack v2.4.1 introduced the [postgres_fields_sort_option](https://github.com/activerecord-hackery/ransack/pull/1184) configuration option (an awesome feature by the way!)

That change has the side effect of converting  `scope_or_sort` from an Arel to a string. This triggers certain versions of Rails to fire a warning every time an order is applied if `postgres_fields_sort_option` is set.

This fix ensures that if the ordering is converted to a string when the `NULLS FIRST`/`NULLS LAST` is appended it is then converted back into an Arel before being passed to `relation.order`. This silences the warnings because the sql is no longer a string when it is run.

More info on the warning can be found here: https://github.com/rails/rails/issues/32995

### DIFF
--- a/lib/ransack/adapters/active_record/context.rb
+++ b/lib/ransack/adapters/active_record/context.rb
@@ -44,9 +44,9 @@ module Ransack
               else
                 case Ransack.options[:postgres_fields_sort_option]
                 when :nulls_first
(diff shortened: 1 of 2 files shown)


Polidoro:
I need to double check that this is reproducible, so closing this until I do.

PR #1209, opened 2021-02-24
Drop Ruby 2.6 from cronjob as Rails 7 requires Ruby 2.7+

yahonda:
This commit addresses these errors at cronjob.
https://github.com/activerecord-hackery/ransack/runs/1966021652?check_suite_focus=true

```
activesupport-7.0.0.alpha requires ruby version >= 2.7.0, which is incompatible
with the current version, ruby 2.6.6p146
Error: Process completed with exit code 5.
```

Refer
https://github.com/rails/rails/commit/6487836af8f50648a9b30ce61864c827132e5592

### DIFF
--- a/.github/workflows/cronjob.yml
+++ b/.github/workflows/cronjob.yml
@@ -13,7 +13,6 @@ jobs:
         ruby:
           - 3.0.0
           - 2.7.2
-          - 2.6.6
     env:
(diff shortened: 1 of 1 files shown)


seanfcarroll:
Awesome, thank you so much for the super-quick response @yahonda ! 🚀

PR #1211, opened 2021-02-24
CI against Rails 6.1.2

yahonda:
- "Rails 6.1.2 has been released!"
https://weblog.rubyonrails.org/2021/2/9/Rails-6-1-2-has-been-released/


yahonda:
My bad. The target version should be 6.1.3.

PR #1218, opened 2021-03-31
Fix :default option in Translate.attribute method

coreyaus:
At the moment it's possible to pass a default translation (or translation key) to the Ransack::Translate.attribute method, however it is never actually accessed given the existing fallback using interpolated attribute/predicate names. Those interpolated strings are either "%{attributes}" or "%{attributes} %{predicate}", which never return nil. Any value passed to options[:default] is added to the end of the defaults array (after those never-nil interpolated strings), and as a result the default that is specifically supplied by the user of the gem is never actually accessed or output. This pull request is aimed at resolving that (ever-so-slight) bug


seanfcarroll:
Looks good, thank you @coreyaus

PR #1223, opened 2021-05-06
Enable sort using alias

waldyr:
Duplicates #766 and #1084 but this one have tests. All credit should be given to the PR authors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -467,6 +467,25 @@ query parameters in your URLs.
 <% end %>
 ```
 
+You can also use `ransack_alias` for sorting.
+
+```ruby
+class Post < ActiveRecord::Base
+  belongs_to :author
+
+  # Abbreviate :author_first_name to :author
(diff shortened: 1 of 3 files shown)


seanfcarroll:
This looks great @waldyr, any chance to add a note to the documentation?

waldyr:
Sure. I just don't know what to add 😆 

The example [here](https://github.com/activerecord-hackery/ransack#ransack-aliases) uses `_or_` so it won't really work with sorting. Should I add another alias on the example and explain that you can use with sort_link or maybe you have something else in mind.

seanfcarroll:
> Should I add another alias on the example and explain that you can use with sort_link

@waldyr this would cover it, thank you.

waldyr:
@seanfcarroll done. Let me know if it's enough

seanfcarroll:
@waldyr that is great, thank you. I'll merge it once the CI has completed.

PR #1224, opened 2021-05-09
params_on_build_condition

yan13to:


deivid-rodriguez:
Could you explain what you're trying to achieve and add some tests? Thanks!

yan13to:
Sorry, i made a mistake, its for my project only, need some default values on build condition, so please delete it, i already merged it in my fork. Thank You very much

PR #1225, opened 2021-05-12
Allow scopes that define string SQL joins

mhssmnn:
Allows scopes to add a string join for more complex joins. These changes fix an issue when using a string join in combination with a **negated condition on a relation**:

```
NoMethodError:
    undefined method `children' for nil:NilClass
    # ./lib/ransack/adapters/active_record/context.rb:143:in `block in remove_association'
```

because this line was acting on a string:

```ruby
@object.joins_values.delete_if { |jd|
    jd.instance_variable_get(:@join_root).children.map(&:object_id) == [association.object_id]
}
```


seanfcarroll:
This is a nice feature @mhssmnn, thank you 🎉 
Would it be possible to add a note to the README ?

PR #1226, opened 2021-05-19
allow ransack to treat nulls as always first or last

mollerhoj:
Current sort options are strangely determined by ASC, DESC order. This fix allows to circumvent that, without breaking backwards compatibility.


seanfcarroll:
Thank you for your contribution @mollerhoj

PR #1241, opened 2021-07-06
No need to skip the test because Rails 6.0.4 is out

yahonda:
This commit partially reverts #1171

Rails 6.0.4 has been released including the commit below.
https://github.com/rails/rails/commit/f9ba52477ca288e7effa5f6794ae3df3f4e982bc


seanfcarroll:
Many thanks @yahonda 🎉

PR #1249, opened 2021-09-10
Squeel should no longer be recommended

Schwad:
See background: https://github.com/activerecord-hackery/squeel/pull/428


seanfcarroll (commented):
@Schwad thanks for the PR, I also removed the reference to Metasearch.

PR #1255, opened 2021-10-06
Added support for matches_regexp

lloydwatkin:
This is a commit from our old fork, it adds regular expression support to `ransack`.

Dropping it here so the contribution isn't lost. I'll try and look into making all the necessary changes soon - original change on our fork wasn't by myself.


scarroll32:
Looks like a great contribution @lloydwatkin, but can you fix the specs?

Could you also add some new specs to test the regex?

scarroll32:
> [!NOTE]

Thanks @lloydwatkin for dropping this here so it wasn't lost — closing it, since it isn't mergeable in its current form and you'd flagged as much yourself.

As it stands the change only adds `'matches_regexp'` to the list of Arel predicate names; nothing registers a Ransack predicate for it, so `name_matches_regexp` isn't recognised. The accompanying spec exercises a different name (`name_matches_regex`) and asserts `LIKE` / `REGEX` SQL that no adapter produces. It fails against `main`.

A real implementation is worth having, but it's adapter-specific in a way the rest of Ransack isn't: Arel's `matches_regexp` renders `~` on PostgreSQL and `REGEXP` on MySQL, and SQLite has no `REGEXP` function unless the application registers one. It would need a predicate registered via `add_predicate` (plus `does_not_match_regexp` and the case-insensitive variants), per-adapter SQL expectations on the PostgreSQL and MySQL CI legs, an explicit stance on SQLite, and docs in Search Matchers. If anyone wants to pick that up, a fresh PR against current `main` is the place — there's no open issue for it, so feel free to open one to gauge interest first.

PR #1113, opened 2020-04-03
Polyamorous module

varyonic:
Resumes #1105  Closes #1099  Fixes #1081 ?

PR #1275, opened 2022-03-04
Fix issue with self joins and includes

### LINKED ISSUE #1271: Joins problem
Please execute below test script. Agent account of account record should be nil but it's not. And the SQL use trade_accounts_accounts table name twice. `"trade_accounts_accounts"."id" AS t1_r0` and `"trade_accounts_accounts"."id" AS t2_r0`

```ruby
SQL (0.3ms)  SELECT "accounts"."id" AS t0_r0, "accounts"."agent_account_id" AS t0_r1, "accounts"."trade_account_id" AS t0_r2, "trade_accounts_accounts"."id" AS t1_r0, "trade_accounts_accounts"."agent_account_id" AS t1_r1, "trade_accounts_accounts"."trade_account_id" AS t1_r2, "trade_accounts_accounts"."id" AS t2_r0, "trade_accounts_accounts"."agent_account_id" AS t2_r1, "trade_accounts_accounts"."trade_account_id" AS t2_r2 FROM "accounts" INNER JOIN "accounts" "trade_accounts_accounts" ON "trade_accounts_accounts"."id" = "accounts"."trade_account_id" ORDER BY "accounts"."id" ASC LIMIT ?  [["LIMIT", 1]]
```

ruby version: 3.0.2

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  # Activate the gem you are reporting the issue against.
  gem "rails", "6.1.4.6"
  gem "sqlite3"
  gem "ransack", "~> 2.5", require: false
end

require "active_record"
require "minitest/autorun"
require "logger"
require "ransack"

# This connection will do for database-independent bug reports.
ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")

  create_table :accounts, force: true do |t|
    t.belongs_to :agent_account
    t.belongs_to :trade_account
  end
end

class Account < ActiveRecord::Base
  belongs_to :agent_account, class_name: "Account"
  belongs_to :trade_account, class_name: "Account"
end

class BugTest < Minitest::Test
  def test_joins_with_includes
    trade_account = Account.create!
    Account.create(trade_account: trade_account)

    accounts = Account.joins(:trade_account).includes(:trade_account, :agent_account)
    account = accounts.first

    assert_equal nil, account.agent_account
  end
end

```

deivid-rodriguez:
Our polyamorous monkeypatches were doing too much. Removing our custom equality and letting ActiveRecord figure it out seems to do the trick and fix this.

On some Rails & DB adapters this seems to break a polyamorous spec. However, this spec just tests an implementation detail but it's unclear whether the spec breaking has any actual bad side effect in real like, so I removed it.

Fixes #1271.

### DIFF
--- a/lib/polyamorous/activerecord_5.2_ruby_2/join_association.rb
+++ b/lib/polyamorous/activerecord_5.2_ruby_2/join_association.rb
@@ -16,9 +16,5 @@ module Polyamorous
         super(reflection, children)
       end
     end
-
-    def ==(other)
-      base_klass == other.base_klass
-    end
   end
 end
--- a/lib/polyamorous/activerecord_6.1_ruby_2/join_association.rb
+++ b/lib/polyamorous/activerecord_6.1_ruby_2/join_association.rb
@@ -66,9 +66,5 @@ module Polyamorous
 
       joins
     end
-
-    def ==(other)
-      base_klass == other.base_klass
-    end
   end
 end
--- a/spec/polyamorous/activerecord_compatibility_spec.rb
+++ b/spec/polyamorous/activerecord_compatibility_spec.rb
@@ -1,0 +1,15 @@
+require 'spec_helper'
+
+module Polyamorous
+  describe "ActiveRecord Compatibility" do
+    it 'works with self joins and includes' do
+      trade_account = Account.create!
+      Account.create!(trade_account: trade_account)
+
+      accounts = Account.joins(:trade_account).includes(:trade_account, :agent_account)
+      account = accounts.first
+
+      expect(account.agent_account).to be_nil
+    end
+  end
+end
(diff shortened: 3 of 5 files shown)


deivid-rodriguez:
I investigated a bit where this comes from.

These equality methods and the corresponding spec were correctly removed at https://github.com/activerecord-hackery/ransack/pull/946, to keep up to date with a change in active record. However, they were incorrectly added back at https://github.com/activerecord-hackery/ransack/pull/1113 when polyamorous code was included with ransack. In particular by https://github.com/activerecord-hackery/ransack/pull/1113/commits/ca51ed8a1749fb39e472fec72cc6f938798e7f0c and https://github.com/activerecord-hackery/ransack/pull/1113/commits/ca51ed8a1749fb39e472fec72cc6f938798e7f0c, which suggests that this is a regression only present in Ransack 2.4.0 or newer.

PR #1287, opened 2022-03-21
Use `bundle add` in README installation instructions

glaucocustodio:
As per https://github.com/rubygems/rubygems/pull/5337, we can simplify the steps of adding a gem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,11 +23,10 @@ Run to install:
 bundle add ransack
 ```
 
-If you would like to use the latest updates (recommended), use the `master`
-branch:
+If you would like to use the latest updates, use the `master` branch:
 
 ```ruby
-gem 'ransack', github: 'activerecord-hackery/ransack'
+bundle add ransack --github "activerecord-hackery/ransack"
 ```
 
 ## Issues tracker
--- a/README.md
+++ b/README.md
@@ -17,10 +17,10 @@ or controller layer, you're probably not looking for Ransack.
 
 Ransack is supported for Rails 7.0, 6.x on Ruby 2.6.6 and later.
 
-In your Gemfile, for the last officially released gem:
+Run to install:
 
 ```ruby
-gem 'ransack'
+bundle add ransack
 ```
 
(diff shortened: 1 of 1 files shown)


deivid-rodriguez (commented):
In this case, since we're are also providing instructions to install the unreleased gem from the master branch directly, I think we should either migrate both alternatives to use `bundle add`, or leave things as they are.

What do you think?

glaucocustodio:
I can't get `bundle add ransack --git=git@github.com:activerecord-hackery/ransack.git` working, I get:

```sh
[!] There was an error parsing `injected gems`: You cannot specify the same gem twice with different version requirements.
You specified: ransack (~> 2.4.2) and ransack (>= 0). Gem already added. Bundler cannot continue.

 #  from injected gems:1
 #  -------------------------------------------
 >  gem "ransack", ">= 0", :git => "git@github.com:activerecord-hackery/ransack.git"
```

glaucocustodio:
Anyway I think is valid having `bundle add` to let people know about this :)

deivid-rodriguez:
Mmmm, maybe a `bundle add` issue? If we don't migrate both instructions, I think it makes things more confusing that they currently are to be honest.

glaucocustodio:
Why do we recommend people using the master version? This is odd, as any breaking change would affect people pointing to it..

README.md
```diff
@@ -17,17 +17,16 @@ or controller layer, you're probably not looking for Ransack.
 
 Ransack is supported for Rails 7.0, 6.x on Ruby 2.6.6 and later.
 
-In your Gemfile, for the last officially released gem:
+Run to install:
```
deivid-rodriguez, line 20: Can we say that the command below also adds the gem to the `Gemfile` automatically? Maybe something like

```suggestion
To install `ransack` and add it to your Gemfile, run
```

deivid-rodriguez (commented):
Minor wording issue, but looks good to me now.

PR #1289, opened 2022-03-24
Deprecate passing two trailing hashes to `sort_link`

deivid-rodriguez:
Pass a single options hash instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


dmitrue:
Hi, with this merged how to correctly pass data: { turbo_action: :advance } to `sort_link` without appending it to params string now?
`items?data-turbo-action=advance&q%5Bs%5D=id+asc"`

deivid-rodriguez:
Hi, that seems like an oversight on my part, I think `data` needs to be special cased here like we did with `class:`. I'll work on it.

PR #1301, opened 2022-04-07
Don't pass `data` attribute appended to query params

deivid-rodriguez:
Fixes https://github.com/activerecord-hackery/ransack/pull/1289#issuecomment-1084225840.

@dmitrue Can you verify whether this fixes your issue?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


dmitrue:
@deivid-rodriguez Yes, it fixes issue with passing data attributes for me, thank you!

PR #1205, opened 2021-02-08
Bump the Rails version to 7

yahonda:
This pull request addresses the `LoadError:
cannot load such file -- polyamorous/activerecord_7.0_ruby_2/join_association`
reported at https://github.com/activerecord-hackery/ransack/runs/1835251570?check_suite_focus=true

Refer https://github.com/rails/rails/commit/1b455e2e9d6937d4107e19cb32e2f98aa08886b9
Fix #1204

PR #1309, opened 2022-04-12
Allow Ransack runs test against Rails main branch aka 7.1.0.alpha

yahonda:
Refer to #1192 and #1205 for the similar change to support 6.2.0.alpha
and 7.0.0.alpha

This commit should make cronjob CI green.


yahonda:
We can see the "cronjob" status via https://github.com/activerecord-hackery/ransack/actions/workflows/cronjob.yml

deivid-rodriguez (approved):
Thanks!

yahonda:
The cronjob is green now. https://github.com/activerecord-hackery/ransack/actions/runs/2158246413

PR #1310, opened 2022-04-12
Tidy up CI

yahonda:
This pull request updates CI as follows.

* Update Rails version to the latest Rails 7.0.2, 6.1.5
* Remove CI against 6-0-stable branch because Rails 6.0.z only accepts severe security issues
* Add Ruby 3.1.1, removed Ruby 2.6.7 from CI

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,11 +11,11 @@ jobs:
       fail-fast: false
       matrix:
         rails:
-          - v7.0.0
-          - v6.1.4
+          - v7.0.2
+          - v6.1.5
           - v6.0.4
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
We need to drop support for Ruby 2.6 if we stop testing with it. It's no longer supported since the end of March, as per https://www.ruby-lang.org/en/news/2021/11/24/ruby-2-6-9-released/, so it should be fine. Can you create a separate PR to do that, since it's a user visible change?

yahonda:
Sure.

deivid-rodriguez (approved):
👍

PR #1321, opened 2022-05-10
Fix links to Bug Report Templates

ioki-klaus:
I noticed that the bug report template links were broken, this links to the current version


deivid-rodriguez (approved):
Thank you!

PR #1325, opened 2022-05-19
Cast PostgreSQL's timestamptz columns to time

lucasfais:
Before Rails 7, PostgreSQL's timestamptz columns were handled as `datetime`, but now `type` returns `:timestamptz`:

```
irb(main):181:0> User.columns_hash['created_at'].type
=> :timestamptz
```

I tested this PR's change in my production app, it's working fine, but let me know if this is not the correct/complete fix for the issue.

### DIFF
--- a/lib/ransack/nodes/value.rb
+++ b/lib/ransack/nodes/value.rb
@@ -26,7 +26,7 @@ module Ransack
         case type
         when :date
           cast_to_date(value)
-        when :datetime, :timestamp, :time
+        when :datetime, :timestamp, :time, :timestamptz
           cast_to_time(value)
         when :boolean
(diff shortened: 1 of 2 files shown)


scarroll32:
@lucasfais thank you for this PR.

Could you please add a couple of tests?

lucasfais:
@scarroll32 Sure. `Ransack::Nodes::Value` had no specs, so I added basic specs for all types. `:timestamptz` specially is on line 28. Thanks!

deivid-rodriguez (approved):
Makes sense to me!

scarroll32:
Thank you @lucasfais !

PR #1332, opened 2022-05-24
Fix broken documentation link

cpgo:
The current link points to a github wiki that apparently is not there anymore.


deivid-rodriguez (approved):
Thank you!

Issue #1344, opened 2022-06-12
Remove dead MongoDB code


scarroll32:
Remove dead MongoDB code. 

See also https://github.com/activerecord-hackery/ransack/issues/1346

PR #1345, opened 2022-06-12
Remove dead MongoDB code

scarroll32:
Closes https://github.com/activerecord-hackery/ransack/issues/1344

### DIFF
--- a/lib/ransack/adapters.rb
+++ b/lib/ransack/adapters.rb
@@ -2,17 +2,7 @@ module Ransack
   module Adapters
 
     def self.object_mapper
-      @object_mapper ||= instantiate_object_mapper
-    end
-
-    def self.instantiate_object_mapper
-      if defined?(::ActiveRecord::Base)
-        ActiveRecordAdapter.new
-      elsif defined?(::Mongoid)
-        MongoidAdapter.new
-      else
-        raise "Unsupported adapter"
-      end
+      @object_mapper ||= ActiveRecordAdapter.new
     end
 
     class ActiveRecordAdapter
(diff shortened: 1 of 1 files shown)


scarroll32:
@deivid-rodriguez could you review this change please.

deivid-rodriguez:
Is this used by https://github.com/activerecord-hackery/ransack-mongoid?

If no, then this is good to go as is without any concerns.

If yes, then this seems also fine anyways given that you [already archived the ransack-mongoid repo back in march](https://github.com/activerecord-hackery/ransack-mongoid/commit/e75d8112b37146991f73548523e7fc3621c407ff), but we consider backwards compatibility, right?

scarroll32:
> If yes, then this seems also fine anyways given that you [already archived the ransack-mongoid repo back in march](https://github.com/activerecord-hackery/ransack-mongoid/commit/e75d8112b37146991f73548523e7fc3621c407ff), but we consider backwards compatibility, right?

@deivid-rodriguez yes this is correct, the repo is archived. There was little interest for this feature, and if this is something that might be wanted in future it would probably make sense to write it from the ground-up for Mongo, IMHO.

deivid-rodriguez:
Ok, you know best the history of this. I think to be strict with backwards compatibility we should ship this with a major version, but other than that looks fine.

scarroll32:
> Ok, you know best the history of this. I think to be strict with backwards compatibility we should ship this with a major version, but other than that looks fine.

That's an excellent point @deivid-rodriguez ! 

Should we hold this (and https://github.com/activerecord-hackery/ransack/pull/1348) until the next version bump, and make it 4.0.0? 

We are currently on a 2 week cycle for version bumps?

deivid-rodriguez:
We were but I lost the rhythm a couple of cycles ago... 😅. We can recover it on Thursday and get 4.0.0 out if it seems fine to you?

scarroll32:
> We were but I lost the rhythm a couple of cycles ago... 😅. We can recover it on Thursday and get 4.0.0 out if it seems fine to you?

Works for me!! 🎸

deivid-rodriguez (approved):
Alright let's do it!

Issue #1349, opened 2022-06-12
Gemfile.lock is .gitignored


scarroll32:
For some reason `.gitignore` has [excluded `Gemfile.lock`](https://github.com/activerecord-hackery/ransack/blob/5d4c4d80435c15a73862c6e66e70fef4ab3f7581/.gitignore).

I am not sure if there is a good reason for this (supporting multiple Rails versions perhaps 🤔 )

scarroll32:
Hmmm it does seem like it was excluded intentionally

```
Run bundle install
Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_checker)' has been deprecated. Please call `DidYouMean.correct_error(error_name, spell_checker)' instead.
Fetching https://github.com/rails/rails.git
Fetching gem metadata from https://rubygems.org/.........
Resolving dependencies...
Bundler could not find compatible versions for gem "activesupport":
  In snapshot (Gemfile.lock):
    activesupport (= 6.1.6)
  In Gemfile:
    actionpack was resolved to [7](https://github.com/activerecord-hackery/ransack/runs/6851228005?check_suite_focus=true#step:4:8).0.3, which depends on
      activesupport (= 7.0.3)
    activemodel was resolved to 6.1.6, which depends on
      activesupport (= 6.1.6)
Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.
Error: Process completed with exit code 6.
```

scarroll32:
Closing this.

deivid-rodriguez:
Hi @scarroll32!

I wholeheartedly agree with committing the `Gemfile.lock` file, but I don't want to impose my vision unless there's full agreement with other maintainers.

Indeed it gets a bit tricky when supporting multiple Rails versions, we would need to keep 2 Gemfiles and lockfiles. If this is something you wouldn't mind, I'm happy to tackle it.

PR #1350, opened 2022-06-12
Add Gemfile.lock

scarroll32:
Closes https://github.com/activerecord-hackery/ransack/issues/1349


scarroll32:
Closing without merged as per the comment in https://github.com/activerecord-hackery/ransack/issues/1349

PR #1367, opened 2022-10-25
Add warning about necessary authorization

AmShaegar13:
In my opinion, docs should warn about not authorizing by default. This can cause serious problems when filtering the user model for example. One could easily extract email addresses, password hashes or other sensitive information.


scarroll32:
Thank you @AmShaegar13

AmShaegar13:
My pleasure.

Issue #1373, opened 2022-11-11
NULLS FIRST / LAST for MySQL


stereobooster:
For example `nulls_always_last` in MySQL can be simulated like this

### Option 1

```rb
scope :sort_by_priority_null_asc, -> { order(Arel.sql('priority DESC')) }
scope :sort_by_priority_null_desc, -> { order(Arel.sql('-priority DESC')) }
```

Just an idea. There is undocumented feature for MySQL https://troels.arvin.dk/db/rdbms/#select-order_by. We can use it to implement similar to PostgreSQL behavior. I know this is not the best practice to depend on undocumented feature... just an idea

### Option 2

```rb
scope :sort_by_priority_null_asc, -> { order(Arel.sql('priority IS NULL ASC, priority ASC')) }
scope :sort_by_priority_null_desc, -> { order(Arel.sql('priority IS NULL ASC, priority DESC')) }
```

### Option 3

https://github.com/rails/rails/pull/42245

tttffff:
Hi @stereobooster, I took a bit of a dive into this recently.

- Option 1 will not work in a lot of circumstances. It seems to work great for integer and datetime columns, but not for varchar
- Option 2 seems fine and worked in every circumstance that I tried
- Option 3 will not work for most situations, see the issue that I have raised for more info rails/rails#50078

I've raised a PR with Rails that implements your suggestion in Option 2 in Arel. The PR is linked on the issue.

scarroll32:
> [!NOTE]
> Comment from **Claude** (Claude Code), acting on behalf of @scarroll32.

This works on 5.0.0: Ransack emits Arel's `nulls_first` / `nulls_last` nodes (#1696) and Arel's MySQL visitor has emulated them since Rails 7.2 via your own rails/rails#50079 (`ORDER BY col IS NULL, col ASC`, your option 2). Since 5.0 requires Active Record 7.2+, `config.fields_sort_option = :nulls_last` applies on MySQL too. The configuration docs still say otherwise and will be corrected. Closing.

PR #1376, opened 2022-11-12
Use built-in methods for nulls_first, nulls_last

stereobooster:
PR for #1373. PR not ready for merge - let's discuss. I didn't rename options or tests for now

**Summary**:
- Works for all databases except MySQL in Rails 7
- Works for PostgreSQL in Rails 6

### DIFF
--- a/lib/ransack/adapters/active_record/context.rb
+++ b/lib/ransack/adapters/active_record/context.rb
@@ -44,13 +44,13 @@ module Ransack
               else
                 case Ransack.options[:postgres_fields_sort_option]
                 when :nulls_first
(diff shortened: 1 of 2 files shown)


scarroll32:
Related to https://github.com/activerecord-hackery/ransack/pull/1184

deivid-rodriguez:
@stereobooster Makes sense to me to use built-in Rails methods, better than the other two options you presented at #1373. Can you do the necessary renamings/test fixes?

stereobooster:
@deivid-rodriguez yes I can. The problem is that it is kind of braking change. What shall I do about it?

deivid-rodriguez:
Can you explain the breaking change more? Is it because some supported Rails versions don't yet have these?

stereobooster:
Option is called `postgres_fields_sort_option`, but it would work for other databases as well? Shall I leave it as is?

PR #1384, opened 2022-12-19
Add support for default predicates

p8:
Currently predicates are required, otherwise the conditions are ignored or they will raise an error.
When using Ransack for filtering in json requests, it would be nice to allow an empty predicate to equal to true.

So instead of having to explicitly pass '*_eq':

    {
      product_id_eq: 'product-1',
      type_eq:       'admin'
    }

we can do the following:

    {
      product_id: 'product-1',
      type:       'admin'
    }

This adds a `:default_predicate` configuration option that will be used in case of a missing predicate.

I'm not sure anyone would use a different default, so maybe the config option should be something like: `:default_to_eq_predicate`.

### DIFF
--- a/lib/ransack/configuration.rb
+++ b/lib/ransack/configuration.rb
@@ -101,6 +101,19 @@ module Ransack
       self.options[:ignore_unknown_conditions] = boolean
     end
 
+    # By default Ransack ignores empty predicates. Ransack can also fallback to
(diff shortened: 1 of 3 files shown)


scarroll32:
Thank you @p8 🚀

p8:
Thanks @scarroll32 !

PR #1389, opened 2023-01-05
Fix required Ruby and Rails version

tagliala:
The gemspec states that the minimum required ruby version is 2.7 and the minimum required AR/AS versions are 6.1.5

This commit fixes the readme accordingly

[ci skip]


scarroll32:
Thank you @tagliala

PR #1391, opened 2023-01-11
Prevent changing host through params

AndersGM:
Currently, it is possible to change the host simply by passing a host through params. This way, visiting ```example.com/people?host=some_other_domain``` would then have sort_links with ```http://some_other_domain/people..```.


AndersGM:
@deivid-rodriguez I am not allowed to request a review from you, but maybe you could take a look at this? :-)

deivid-rodriguez (approved):
Thanks so much!

PR #1422, opened 2023-05-18
Add namespaced example of Polymorphic search

craigmcnamara:
It's not obvious how to search for namespaced polymorphic models. This addresses that with an example.


scarroll32:
Thank you @craigmcnamara

PR #1430, opened 2023-07-06
Allow nested groupings (longhand) to be passed

mhssmnn:
Currently we can pass either `{ g: [{ name_eq: "val" }] }` or `{ groupings: [{ name_eq: "val" }] }` to `search` and it behaves as expected. 

However if we nest these values then the **longhand** values either raise, or get ignored:

```
> Model.search({ groupings: [{ groupings: [{ name_eq: "val" }] }] })
> #<ArgumentError: No valid predicate for groupings>
```
but
```
Model.search({ groupings: [{ g: [{ name_eq: "val" }] }])
```
performs as expected.

This small change updates the regex to check for the longhand values too.

### DIFF
--- a/lib/ransack/nodes/grouping.rb
+++ b/lib/ransack/nodes/grouping.rb
@@ -149,7 +149,7 @@ module Ransack
       def build(params)
         params.with_indifferent_access.each do |key, value|
           case key
-          when /^(g|c|m)$/
+          when /^(g|c|m|groupings|conditions|combinator)$/
             self.send("#{key}=", value)
           else
             write_attribute(key.to_s, value)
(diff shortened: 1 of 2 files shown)


scarroll32:
> [!NOTE]

Thanks @mhssmnn, and apologies for the long wait on a three-character fix.

The asymmetry is clearly a bug: `groupings:` and `conditions:` are accepted at the top level but only the `g`/`c`/`m` abbreviations were recognised when nested, so `{ groupings: [{ groupings: [...] }] }` fell through to `write_attribute` and raised `No valid predicate for groupings`. Confirmed `groupings=`, `conditions=` and `combinator=` all exist on `Grouping`, so the widened regex dispatches correctly.

Verified locally merged into current `main` — Rails 8.1.3 / Ruby 3.4.9 / SQLite: **527 examples, 0 failures, 1 pending**.

Merging with an admin override: the branch predates #1667, so the Rails 8.1 / Ruby 3.3–4.0 required contexts can't report on it. All 24 checks it does define are green.

PR #1439, opened 2023-09-25
Fix builds for Rails 7.1.0

yuki24:
Right now, Ransack fails to execute on Rails 7.1.0beta1 with the error below:

```
  1) Ransack::Search#method_missing allows chaining to access nested conditions
     Failure/Error: table        = attr.arel_attribute.relation.table_name

     NoMethodError:
       undefined method `table_name' for #<Arel::Table:...>
```

This is because the `table_name` method has been dropped as part of this commit https://github.com/rails/rails/commit/1d98bc563a8a7faf26238eaaa54a3257a95d644d, which has made Rails imcompatible with Ransack.

This PR adds a check for the existence of the `table_name` method to restore the compatibility.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,6 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         rails:
+          - v7.1.0.rc2
           - v7.0.3
           - v6.1.6
         ruby:
@@ -38,6 +39,7 @@ jobs:
       fail-fast: false
       matrix:
         rails:
+          - v7.1.0.rc2
           - v7.0.3
           - v6.1.6
         ruby:
@@ -72,6 +74,7 @@ jobs:
       fail-fast: false
       matrix:
         rails:
+          - v7.1.0.rc2
           - v7.0.3
           - v6.1.6
         ruby:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         rails:
-          - v7.1.0.rc2
+          - v7.1.0
           - v7.0.3
           - v6.1.6
         ruby:
@@ -39,7 +39,7 @@ jobs:
       fail-fast: false
       matrix:
         rails:
-          - v7.1.0.rc2
+          - v7.1.0
           - v7.0.3
           - v6.1.6
         ruby:
@@ -74,7 +74,7 @@ jobs:
       fail-fast: false
       matrix:
         rails:
-          - v7.1.0.rc2
+          - v7.1.0
           - v7.0.3
           - v6.1.6
         ruby:
--- a/lib/ransack/adapters/active_record/context.rb
+++ b/lib/ransack/adapters/active_record/context.rb
@@ -12,8 +12,9 @@ module Ransack
 
         def type_for(attr)
           return nil unless attr && attr.valid?
+          relation     = attr.arel_attribute.relation
           name         = attr.arel_attribute.name.to_s
-          table        = attr.arel_attribute.relation.table_name
+          table        = relation.respond_to?(:table_name) ? relation.table_name : relation.name
           schema_cache = self.klass.connection.schema_cache
           unless schema_cache.send(:data_source_exists?, table)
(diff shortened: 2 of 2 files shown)


jlupox:
Thanks @yuki24.
It doesn't work for me. I have to change to:

`table        = relation.respond_to?(:table_name) ? relation.try(:table_name) : relation.name`

yuki24:
@jlupox Thanks for pointing that out! It was embarrassingly broken 😅 But it should be fixed now.

engineersmnky:
I would recommend just changing this to 
```
table        = attr.arel_attribute.relation.name
```
Checking for the existence of `table_name` is unnecessary overhead as `Arel::Table#name` has always existed.

yuki24:
@engineersmnky That didn't work, as it's not always the `Arel::Table` object but could be `Arel::TableAlias` object, in which case there are many failing tests:

```
  18) Ransack::Search#method_missing allows chaining to access nested conditions
      Failure/Error: raise "No table named #{table} exists."

      RuntimeError:
        No table named children_people exists.
      # ./lib/ransack/adapters/active_record/context.rb:20:in `type_for'
      # ./lib/ransack/nodes/attribute.rb:35:in `type'
      # ./lib/ransack/nodes/condition.rb:266:in `default_type'
      # ./lib/ransack/nodes/condition.rb:25:in `extract'
      # ./lib/ransack/nodes/grouping.rb:175:in `write_attribute'
      # ./lib/ransack/nodes/grouping.rb:155:in `block in build'
      # ./lib/ransack/nodes/grouping.rb:150:in `each'
      # ./lib/ransack/nodes/grouping.rb:150:in `build'
      # ./lib/ransack/nodes/grouping.rb:146:in `new_grouping'
      # ./lib/ransack/nodes/grouping.rb:95:in `block in groupings='
      # ./lib/ransack/nodes/grouping.rb:94:in `each'
      # ./lib/ransack/nodes/grouping.rb:94:in `groupings='
      # ./lib/ransack/search.rb:106:in `method_missing'
      # ./spec/ransack/search_spec.rb:687:in `block (3 levels) in <module:Ransack>'
```

engineersmnky:
Sorry I missed that because while `Arel::TableAlias` defines name it refers to the wrong side.

In that case I agree the `respond_to` check is needed but the `try` is superfluous.

yuki24:
@engineersmnky The `try` call is not there any more. It would have been greatly appreciated if you had read the actual code in the PR.

yuki24:
Rails 7.1 release is right around the corner, and it would be great if we could get this merged. The tests are all green and the Code Climate check seems trivial (only 1 point of complexity check).

jayshepherd:
> Rails 7.1 release is right around the corner, and it would be great if we could get this merged. The tests are all green and the Code Climate check seems trivial (only 1 point of complexity check).

Looks like it was this morning..  Mind bumping to 7.1.0?

yuki24:
@jayshepherd The world moves too fast! Ju updated the CI config and this should be good now.

davidwessman:
Is there something we can help with to move this along? Testing?

n-rodriguez:
Hi there! Any news?

deivid-rodriguez:
Hello! I'll try to merge and get a release out this week, thanks for the patience!

deivid-rodriguez:
Hei, I was expecting to release this week as I said, but I realized I'd like to give @scarroll32 a chance to validate what's going in, and leave whatever feedback he has. So I'd let this PR sit for a couple of days in main before releasing.

Apologies if I created some false expectations there! New ETA around the middle of next week unless Sean gets to it earlier.

@scarroll32 I left a release draft ready in case you want to have a look. There's not much to be released, really.

scarroll32:
Looks good to me, thank you @deivid-rodriguez and @yuki24

deivid-rodriguez:
Awesome, I will release now then. By the way, I sneaked #1447 into the release in the last minute and will include it too!

Augustin-Grenne:
I'd like to second @iainbeeston request to backport this to ransack 3 as sadly patching 4 proofed insufficient (see my comment) and I'm also waiting on #1403

PR #1442, opened 2023-10-05
Docs: Updates sorting example

### LINKED ISSUE #1419: Missing info in sorting documentation
Hello, 

There is an important missing info in this documentation page : https://activerecord-hackery.github.io/ransack/getting-started/sorting/ 

I'd like to do the same thing as described in the doc, but... instead of the "sort" field used in the form we see: 

```
      <%= render partial: 'filters/date_title_sort', locals: { f: f } %>
```

So... as I don't know what this partial contains, I feel like I'm missing an important piece here :-)

Could you replace that "render partial" with the actual code? 

Or put the partial code in the documentation? 

Thx!

kinduff:
Closes #1419

Updates the sorting code examples to be vanilla and make it clear how to use sorting.

Easy code review in [this gist](https://gist.github.com/kinduff/e512fb88bdb1be1da8503c7373b6bbb0).

### DIFF
--- a/docs/docs/getting-started/sorting.md
+++ b/docs/docs/getting-started/sorting.md
@@ -10,70 +10,62 @@ title: Sorting
 You can add a form to capture sorting and filtering options together.
 
 ```erb
-<div class="filters" id="filtersSidebar">
-  <header class="filters-header">
-    <div class="filters-header-content">
-      <h3>Filters</h3>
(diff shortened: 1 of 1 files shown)


deivid-rodriguez (approved):
Sounds like a nice improvement! Thanks!

PR #1445, opened 2023-10-13
Ransack for edge rails

tvongaza:
Stop copying and pasting directories, instead load the last known good version and try to use that


lib/polyamorous/polyamorous.rb
```diff
@@ -11,9 +11,31 @@ module Polyamorous
...
+  version_numbers = matching_dirs.map do |dir|
+    dir.match(/activerecord_(\d+\.\d+)_ruby_2/)
+  end.compact.map { |m| m[1].split('.').map(&:to_i) } # Split major and minor into integers for easy comparison
+
+  raise "No compatible directories found" if version_numbers.empty?
+
+  # 3. Compare them to the required version and find the closest match
+  sorted_versions = version_numbers.sort_by { |major, minor| [major, minor] }.reverse
+  compatible_version = sorted_versions.find { |major, minor| major <= ar_version_major && minor <= ar_version_minor }
+  compatible_version ||= sorted_versions.last # default to the latest version if none are below the requested version
```
tvongaza, line 33:
```suggestion
  compatible_version ||= sorted_versions.first # default to the latest version if none are below the requested version
```

deivid-rodriguez:
Interesting!

I think the reason this code existed is that we needed different patches for the different versions that we supported. Sounds like things are more stable for us now and in our current support matrix these patches are exact duplicates of each other.

If that's the case, can we instead rename the folder to "activerecord", and always load that, without any magic?

tvongaza:
Good idea, I'll whip up a follow up PR trying to implement the change to simplify things.

deivid-rodriguez:
I just noticed we may actually need your solution or something similar given #1447.

scarroll32:
Nice refactor thanks @tvongaza

scarroll32:
@deivid-rodriguez any idea why the CI is getting stuck? I've re-run it but it stopping in the same place.

Screenshot 2023-10-21 at 12 32 39

scarroll32:
> Good idea, I'll whip up a follow up PR trying to implement the change to simplify things.

Thanks @tvongaza or feel free to push directly to this PR. We can probably merge it as-is, but it would be cleaner with the approach @deivid-rodriguez suggested.

scarroll32:
@tvongaza could you please rebase this?

> This branch is out-of-date with the base branch

deivid-rodriguez:
I don't think we can merge this as is after #1447, since we now have a Rails 6.1 specific monkey patch again.

What I think we can do is, have a main folder named "activerecord", and load that if a folder specific to current rails is not found. How about that?

tvongaza:
Ok, updated the branch with the latest changes. I did not see a 6.1 specific monkey patch, as again 7.0 and 7.1 both just call the 6.1 code. I moved it all to an `activerecord` directory and did away with the specific version loading. I'm always a fan of removing code if it isn't needed, so I removed all the version specific loading, which, IMHO, can be revisited if it is ever needed again.

deivid-rodriguez:
https://github.com/activerecord-hackery/ransack/pull/1447/files changes a 6.1 specific monkey patch so this files should not be identical.

tvongaza:
> https://github.com/activerecord-hackery/ransack/pull/1447/files changes a 6.1 specific monkey patch so these files should not be identical.

But the 7.0 & 7.1 files just point to the 6.1 files - or am I missing something obvious?

They all look like this:

https://github.com/activerecord-hackery/ransack/blob/main/lib/polyamorous/activerecord_7.0_ruby_2/join_association.rb#L1
```require 'polyamorous/activerecord_6.1_ruby_2/join_association'```

**Comment by tvongaza** [comment]
> Can you remove the last commit though? I don't want to run CI in all PRs against a moving target like Rails edge, too unstable. We already run CI against Rails edge on a daily schedule.

Oh nice, did realize the gem had a daily edge rails CI run. Commit removed.

**Comment by deivid-rodriguez** [comment]
Ok, let's merge this. Thanks so much @tvongaza!

**Comment by dalibor** [comment]
Hello - could you please release 4.1.1 (?) with this fix?

[4.1.0](https://rubygems.org/gems/ransack/versions/4.1.0) fails with the following error, but activerecord-hackery's `main` seems to work well (testing with active admin).
```
`require': cannot load such file -- polyamorous/activerecord_7.2_ruby_2/join_association (LoadError)
```

Thanks!

scarroll32:
> Hello - could you please release 4.1.1 (?) with this fix?
> 
> [4.1.0](https://rubygems.org/gems/ransack/versions/4.1.0) fails with the following error, but ransack `main` seems to work well (testing with active admin).
> 
> ```
> `require': cannot load such file -- polyamorous/activerecord_7.2_ruby_2/join_association (LoadError)
> ```
> 
> Thanks!

4.1.1 has been released

CC @deivid-rodriguez

Issue #1253, opened 2021-09-28
Ransack broke ActiveRecord sql when same table and different "belongs_to" joining (even though not use ransack search)


kei-p:
I just updated the ransack gem from 2.3.2 to 2.4.2 and found a puzzling bug.
This bug is lost "JOIN" query partially when same table and different "belongs_to" joining query even though not use ransack search method.
I found this bug caused by ransack 2.4.2 ( maybe 2.4.x)  by below example code and results.

## Results
| rails      | ransack | test   |
| ---------- | ------- | ------ |
| 6-1-stable | v2.4.2   | Failed |
| 6-0-stable | v2.4.2   | Failed |
| 6-1-stable | master(4caf7c2)  | Failed |
| 6-0-stable | v2.3.2   | Pass   |
| 6-1-stable | no use (comment out "ransack")  | Pass   |

## Example code
```
unless File.exist?('Gemfile')
  File.write('Gemfile', <<-GEMFILE)
    source 'https://rubygems.org'

    # Rails master
    gem 'rails', github: 'rails/rails', branch: '6-1-stable'

    gem 'pg'
    gem 'ransack', github: 'activerecord-hackery/ransack', tag: 'v2.4.2'
  GEMFILE

  system 'bundle install'
end

require 'bundler'
Bundler.setup(:default)

require 'active_record'
require 'minitest/autorun'
require 'logger'
require 'ransack'

# This connection will do for database-independent bug reports.
`psql -c "DROP DATABASE IF EXISTS ransack_bug_report;"`
`psql -c "CREATE DATABASE ransack_bug_report;"`
ActiveRecord::Base.establish_connection(adapter: 'postgresql', database: 'ransack_bug_report')
ActiveRecord::Base.logger = Logger.new(STDOUT)

# Display versions.
message = "Running test case with Ruby #{RUBY_VERSION}, Active Record #{
  ::ActiveRecord::VERSION::STRING}, Arel #{Arel::VERSION} and #{
  ::ActiveRecord::Base.connection.adapter_name}"
line = '=' * message.length
puts line, message, line

ActiveRecord::Schema.define do
  create_table :tasks, force: true do |t|
    t.bigint :user_id
    t.bigint :reviewer_id
  end

  create_table :users, force: true do |t|
  end
end

class Task < ActiveRecord::Base
  belongs_to :user
  belongs_to :reviewer, class_name: 'User', optional: true
end

class User < ActiveRecord::Base
  has_many :tasks
end

[...]

Issue #1238, opened 2021-07-01
Invalid references to table aliases when joining multiple tables


yan-hoose:
First of all I'd like to thank the creators and maintainers of Ransack - it's been a valuable tool in many of my projects. Projects ranging from hobby stuff to big commercial applications. Thank you for your work!

Now, I ran into an issue where an invalid table alias is being used in the generated SQL. There are other issues currently open here that are similar but slightly different, that is why I created a new issue. In addition, many people reported here that their issue disappeared with an upgrade to Rails 6.1, but this issue persists with 6.1 too.

Consider this test case:

```ruby
require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  #gem 'activerecord', '6.0.4', require: 'active_record'
  #gem 'activerecord', '6.1.4', require: 'active_record'
  gem 'activerecord', github: 'rails/rails', branch: '6-1-stable', require: 'active_record'

  gem 'ransack', github: 'activerecord-hackery/ransack'
  #gem 'ransack', '2.4.2'
  gem 'sqlite3'
end

require 'minitest/autorun'
require 'logger'

ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: ':memory:')
ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  create_table :users, force: true do |t|
  end

  create_table :groups, force: true do |t|
  end

  create_table :groups_users, force: true do |t|
    t.integer :group_id
    t.integer :user_id
  end

  create_table :tickets, force: true do |t|
    t.integer :created_by_id
    t.integer :user_id
  end
end

class User < ActiveRecord::Base
  has_many :groups_users
  has_many :groups, through: :groups_users
end

class GroupsUser < ActiveRecord::Base
  belongs_to :group
  belongs_to :user
end

class Group < ActiveRecord::Base
  has_many :groups_users
  has_many :users, through: :groups_users
end

class Ticket < ActiveRecord::Base
  belongs_to :created_by, class_name: 'User'
  belongs_to :user
end

class BugTest < Minitest::Test
  def test_bug
    q = Ticket.ransack({created_by_groups_users_group_id_eq: 1})

    tickets = q.result.joins('LEFT OUTER JOIN users ON tickets.user_id = users.id')

    puts tickets.to_sql
  end
end
```

The resulting SQL is this:

```sql
SELECT "tickets".*
FROM "tickets"
[...]

Issue #1227, opened 2021-05-20
Invalid table alias when searching has_many through associations


mbiang:
When searching a 'has_many thought' association, a invalid table alias is used in the SQL query if a search parameter for the cousin object is passed in the params has prior to the parameter for the sibling object.

Here is an example:

'
class A < ActiveRecord::Base
  belongs_to :b, optional: true
  has_one :c, through: :b
end

class B < ActiveRecord::Base
  has_many :as
  belongs_to :c
end

class C < ActiveRecord::Base
  has_many :bs, dependent: :destroy
  has_many :as, through: :b
end

# the following call will succeed
A.search({b_name_equals: 'B', c_name_equals: 'C'}).result

# the following call will fail with this postgres error:
# ActiveRecord::StatementInvalid (PG::UndefinedTable: ERROR:  missing FROM-clause entry for table "bs_as")
# LINE 1: ..."."id" = "bs"."c_id" WHERE ("cs"."name" = 'C' AND "bs_as"."n..
A.search({c_name_equals: 'C', b_name_equals: 'B'}).result
`

Apparently, when building the query, it first joins on "b"  in order to join on "cs" to search c.name, and then it it incorrectly assumes the table alias for "bs" as "bs_as", when it should just be "bs".

Ruby Version: 2.7.3
Rails Version: 6.1.3.2
Ransack Version: 2.4.2
Postgres Version: 10.4
Active Admin Version: 2.9.0

Issue #1217, opened 2021-03-29
Incorrect references to table aliases in Rails 6.1


ozzyaaron:
I've gotta start this with thanks so much for Ransack, really the only time it ever gets in the way are during Rails upgrades 😄 

I've seen other issues like the one I'm reporting (https://github.com/activerecord-hackery/ransack/issues/1182, https://github.com/activerecord-hackery/ransack/issues/1153, https://github.com/activerecord-hackery/ransack/issues/1144) but they're on Rails 6.0 or have been fixed in the Rails 6.0 stable branch that will probably be released as Rails 6.0.4. We are upgrading to Rails 6.1 and am hitting table alias issues as part of that. 

* If I use `gem 'rails', github: 'rails/rails', branch: '6-0-stable'` then the below example will work.
* If I use Rails 6.0.3.x then the first 2 assertions work and the last 2 generate a where clause using a table alias that does not exist.
* If I use Rails 6.1.x then the first 2 assertions work and the last 2 generate a where clause using a table alias that does not exist.
* If I use `gem 'rails', github: 'rails/rails', branch: '6-1-stable'` then the first 2 assertions work and the last 2 generate a where clause using a table alias that does not exist.

In general the error raised is

```
ActiveRecord::StatementInvalid: SQLite3::SQLException: no such column: user_applications_activities.name
```
As the table `user_applications_activities` does not exist in the gathering part of the query. 

The only difference between the first 3 assertions and the last 3 are the ordering of the search models & attributes (`user_name_or_user_application_name` vs `user_application_name_or_user_name`).

I added another has relationship (to `somethings`) and started to query on that and it appears that it doesn't break the SQL as badly if that information helps debugging 🤷 

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  # Issue DOES occur with Rails 6.1.x, 6-1-stable & 6.0.3
  gem "activerecord", "~> 6.1.0", require: "active_record"
  # gem "activerecord", github: "rails/rails", branch: "6-1-stable", require: "active_record"
  # gem "activerecord", "~> 6.0.0", require: "active_record"

  # Issue does NOT occur with Rails 6.0-stable (future 6.0.4)
  # gem "activerecord", github: "rails/rails", branch: "6-0-stable", require: "active_record"

  gem "ransack", github: "activerecord-hackery/ransack"
  gem "niceql" # Just makes it nicer to see the SQL
  gem "sqlite3"

  gem "pry"
[...]

PR #1447, opened 2023-10-18
Fix wrong table aliases (#1217)

oneiros:
I encountered the same issue as reported in #1217 (and possibly in a bunch of other issues as well[^1]).

It was correctly pointed out, that a change in ActiveRecord was responsible. https://github.com/rails/rails/commit/10b36e81a357f8d7fa3665630c4d41c057fe59d9 tried to fix a bug with eager loading that resulted from aliasing a joined table that was (then incorrectly) referenced in a WHERE clause.

The solution (as far as I understand it - I never had to look at that code before and cannot say I understand all of what is going on there) was to cache tables already joined and reuse those joins if possible.

The problem in #1217 is caused by ransack prepopulating table information *before* this deduplication code is run. So there is no way to detect that a JOIN clause with an alias, that would otherwise have been used, fell victim to the deduplication.

My "solution" is a poor one. I simply copied the deduplication logic from ActiveRecord. This should give the same results as ActiveRecord would come up with on its own.

A better solution would of course be to somehow get the table information at a later point in time and/or extract the deduplication logic in AR to be able to reuse this more easily. But I suspect the first approach would take a huge refactoring of ransack and I am not 100% sure if it is even possible. Also, I am not convinced that ActiveRecord currently cleans up references to table aliases that are not being used due to the deduplication correctly. So patches for AR itself might be needed anyways.

---

[^1]: The following issues seem like they could be related: #1238, #1253, #1227, #1153, #1144

### DIFF
--- a/lib/polyamorous/activerecord_6.1_ruby_2/join_dependency.rb
+++ b/lib/polyamorous/activerecord_6.1_ruby_2/join_dependency.rb
@@ -56,11 +56,21 @@ module Polyamorous
     private
 
     def table_aliases_for(parent, node)
+      @joined_tables ||= {}
       node.reflection.chain.map { |reflection|
-        alias_tracker.aliased_table_for(reflection.klass.arel_table) do
-          root = reflection == node.reflection
-          name = reflection.alias_candidate(parent.table_name)
-          root ? name : "#{name}_join"
+        table, terminated = @joined_tables[reflection]
+        root = reflection == node.reflection
+
+        if table && (!root || !terminated)
+          @joined_tables[reflection] = [table, true] if root
+          table
+        else
+          table = alias_tracker.aliased_table_for(reflection.klass.arel_table) do
+            name = reflection.alias_candidate(parent.table_name)
+            root ? name : "#{name}_join"
+          end
+          @joined_tables[reflection] ||= [table, root] if join_type == Arel::Nodes::OuterJoin
+          table
         end
       }
     end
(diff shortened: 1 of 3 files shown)


oneiros:
> My "solution" is a poor one. I simply copied the deduplication logic from ActiveRecord.

On second thought: The original implementation of the `table_aliases_for` method that I patched was also just copied from AR. So conceptually my approach is not worse :slightly_smiling_face:

deivid-rodriguez:
Wow, this is so awesome! It's been an unresolved issue for a long time, and biting several people as you point out! Thanks 😍.

I have one question, we even got some regression specs contributed to demonstrate the issue at https://github.com/activerecord-hackery/ransack/pull/1280. Do you think that spec is worth keeping on top of the ones you added, or they are basically duplicates and should be discarded?

oneiros:
Thank you so much for responding so quickly and for the praise :blush: 

> I have one question, we even got some regression specs contributed to demonstrate the issue at https://github.com/activerecord-hackery/ransack/pull/1280. Do you think that spec is worth keeping on top of the ones you added, or they are basically duplicates and should be discarded?

Those seem to be roughly equivalent to what I did (and I wish I had seen them sooner). The main difference seems to be the number of test records created (I used the bare minimum). Other than that it seems to demonstrate the exact same problem and would be redundant. I only had a cursory glance though, I might have missed some nuance there.

deivid-rodriguez:
Awesome, getting this in then, thanks again!

PR #1459, opened 2023-10-27
Update context to support both %i and %w for ransachable_* methods

vimutter:
Currently `ransackable_scopes` allows both, and for some reason the rest expects %w only.


scarroll32:
Thank you @vimutter

PR #1463, opened 2023-11-17
Add NULL FIRST and NULL LAST for most databases

tttffff:
This PR changes how `NULL FIRST` and `NULL LAST` is implemented for Postgresql and also allows it to be used by other DB's. Instead of using SQL fragments, it uses Arel methods.

This is related to #1373, but does not actually fix the issue for MySQL.
I may look into a PR for that at some point in the future. For example, it could be fixed with a conditional on `::ActiveRecord::Base.connection.adapter_name` and SQL fragments for MySQL.

If MySQL users add this configuration, they will get `ActiveRecord::StatementInvalid` errors in some circumstances, in some other other circumstances, they may get unexpected ordering. Becuase of this, I have stated on the documentation that it does not work for MySQL. This has the advantage of just working if Arel is updated to support for MySQL. See rails/rails#50078 for a longer description of how Arel handles null_first/null_last for MySQL at the moment.

### DIFF
--- a/spec/ransack/search_spec.rb
+++ b/spec/ransack/search_spec.rb
@@ -614,19 +614,19 @@ module Ransack
         expect(@s.result.first.id).to eq 1
       end
 
-      it "PG's sort option", if: ::ActiveRecord::Base.connection.adapter_name == "PostgreSQL" do
+      it "fields sort option", if: ::ActiveRecord::Base.connection.adapter_name != "Mysql2" do
         default = Ransack.options.clone
 
         s = Search.new(Person, s: 'name asc')
         expect(s.result.to_sql).to eq "SELECT \"people\".* FROM \"people\" ORDER BY \"people\".\"name\" ASC"
 
-        Ransack.configure { |c| c.postgres_fields_sort_option = :nulls_first }
+        Ransack.configure { |c| c.fields_sort_option = :nulls_first }
         s = Search.new(Person, s: 'name asc')
         expect(s.result.to_sql).to eq "SELECT \"people\".* FROM \"people\" ORDER BY \"people\".\"name\" ASC NULLS FIRST"
         s = Search.new(Person, s: 'name desc')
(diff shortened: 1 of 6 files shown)


spec/ransack/search_spec.rb
```diff
@@ -614,19 +614,19 @@ def remove_quotes_and_backticks(str)
         expect(@s.result.first.id).to eq 1
       end
 
-      it "PG's sort option", if: ::ActiveRecord::Base.connection.adapter_name == "PostgreSQL" do
+      it "fields sort option", if: ::ActiveRecord::Base.connection.adapter_name != "Mysql2" do
```
tttffff, line 617: If Arel does implement working `nulls_first()` and `nulls_last()` for MySQL, or if Ransack is ever updated to implement it, then we will need specific MySQL tests as the generated SQL will be different.

scarroll32:
@copilot can you please fix the conflicts in search_spec.rb

scarroll32:
> [!NOTE]

Thanks @tttffff — replacing the interpolated SQL with `nulls_first`/`nulls_last` is clearly the right move, and your point that it makes the feature backend-agnostic (and future-proof if Arel gains MySQL support) is the strongest argument for it. Rebased as #1696, with you as co-author.

One change: your version renamed `postgres_fields_sort_option` outright. It's a documented public setting, so any initializer using it would hit `NoMethodError` on boot with no migration path. I kept the old name as an alias delegating to the new one, so existing configs keep working while the accurate name becomes the documented one.

> I have stated on the documentation that it does not work for MySQL

Kept and expanded — there's now a "Sorting NULLs" section with a table of the four values and the MySQL caveat stated plainly rather than left to be discovered at runtime.

I also verified it works on SQLite, which the old `postgres_` name implied it didn't:

```ruby
Ransack.configure { |c| c.fields_sort_option = :nulls_first }
Person.ransack(s: 'name asc').result.to_sql
# ... ORDER BY "people"."name" ASC NULLS FIRST
```

Your specs now run on every non-MySQL adapter rather than PostgreSQL only.

Closing in favour of #1696.

Issue #1396, opened 2023-01-23
`ransack!` should fail when given an invalid combinator


kalsan:
Hi and thanks for providing this incredible useful gem. Here's a small change that could make a developer's life even simpler:

Line in question: `User.ransack!(id_eq: 1, first_name_cont: 'Henry', combinator: :or)`

The combinator is invalid because it is of type `Symbol` instead of `String`. Since we are using `ransack!` and not `ransack`, I'd expect this to fail. However, Ransack silently converts this to `'and'`, which is unexpected (in my opinion).

I feel that if the bang method `ransack!` is used, any unexpected input (including to `combinator:`, `m:` etc.), be it the wrong class or the wrong word, should lead to an exception. However, I understand that this is a subjective topic and others might disagree.

Best,
Kalsan

scarroll32:
@kalsan might this be handled by a custom rspec matcher?

kalsan:
@scarroll32 I'm not familiar with rspec, could you elaborate?

kalsan:
@bagp1 thank you very much for addressing this. Much appreciated. This issue can be closed whenever you are ready :-)

Cheers!
Kalsan

PR #1465, opened 2023-12-04
Invalid combinators to raise ArgumentError

bagp1:
Based on [#1396 ](https://github.com/activerecord-hackery/ransack/issues/1396)

The reporter encountered two issues:
1. Passing a combinator value as a symbol doesn't match the expected combinator.
2. Invalid arguments silently misbehavior rather than raise argument errors even when the dangerous method `#ransack!` is called.

I can see why unrecognised combinators default to "AND", it's the more conservative, safe option. But it's very surprising and a potential source of bugs for downstream users. Also the users have intentionally called the dangerous method, so should be error handling as appropriate for their domain. The right thing to do is to raise on invalid args.

However we'd best minimize impact for users currently passing invalid args. The most obvious errors would be rather than passing the expected "or", they might be passing one of "OR", :or and not noticing the fault. We can correct behavior in these common cases by downcasing and converting to a string. Indeed Ransack already does conversion to a string in _some_ pathways, just not as part of a search. Also Anything more exotic is really unrecognisable and an error.

This PR comes with a set of tests describing expected behavior and including a few that would have failed prior to the changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Dangerous searches (e.g. ransack!) will now raise an ArgumentError when passed an unrecognised combinator rather than silently fall back on "AND".
+* Combinator values "OR" and :or are now treated as the expected 'or'.
+
 ## 4.1.0 - 2023-10-23
 
 ### 🚀 Features
--- a/lib/ransack/nodes/condition.rb
+++ b/lib/ransack/nodes/condition.rb
@@ -122,8 +122,9 @@ module Ransack
       end
 
       def combinator=(val)
-        @combinator = Constants::AND_OR.detect { |v| v == val.to_s } || nil
+        super
       end
+
       alias :m= :combinator=
       alias :m :combinator
 
--- a/lib/ransack/nodes/grouping.rb
+++ b/lib/ransack/nodes/grouping.rb
@@ -11,9 +11,13 @@ module Ransack
 
       delegate :each, to: :values
 
+      def combinator=(val)
+        super
+      end
+
       def initialize(context, combinator = nil)
         super(context)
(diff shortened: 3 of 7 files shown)


scarroll32:
> [!NOTE]

Thanks @bagp1 — the premise is right and this is a genuinely valuable fix. Silently falling back to AND on a typo'd combinator means returning the *wrong rows* with no signal, which is the worst failure mode a search library can have. Rebased as #1694, with you as co-author.

One part of it didn't work, and the reason is subtle enough to be worth writing down.

Your normalising writer on `Node` was correct, but it was never reached from the `g:` path. `Grouping` had:

```ruby
attr_accessor :combinator   # line 4 — generates a plain writer
alias :m :combinator
alias :m= :combinator=      # line 6 — binds to THAT writer
...
def combinator=(val)        # line 13 — too late for the alias
  super
end
```

Ruby binds an alias to the method body present **at alias time**, so `m=` kept pointing at the attribute writer generated three lines earlier. Your `def combinator=` applied to `combinator=` but never to `m=` — and `m=` is the path `g: [{ m: 'OR' }]` takes. Checked against your branch:

```ruby
Person.ransack(g: [{ m: 'OR', name_eq: 'a', email_eq: 'b' }]).result.to_sql
# ... WHERE ("people"."name" = 'a' AND "people"."email" = 'b')   ← still AND
```

`Condition` had the ordering right, which is why it looked like it worked. In #1694 `Grouping` exposes only a reader and inherits the writer, so both spellings normalise. I added eight parameterised specs through the `g:` path covering `'or'`, `'OR'`, `'Or'`, `:or`, `:OR` and the `and` forms — four of them fail against `main`.

Also dropped the `CHANGELOG.md` hunk, since since 4.4.0 the changelog comes from GitHub Release notes.

Closing in favour of #1694.

Issue #1467, opened 2023-12-15
NoMethodError: undefined method `value' for #<Arel::SelectManager


spaghetticode:
I'm experiencing this issue starting from Ransack 3.2 which removes the support for Rails 6 and some ancillary code. 

The issue is happening [here](https://github.com/activerecord-hackery/ransack/blob/main/lib/ransack/nodes/condition.rb#L342) and is caused by the fact that in my context the right predicate for a `Arel::Nodes::In` predicate is a `Arel::SelectManager` rather than a `Arel::Nodes::Casted`, so we can't call `#value` on it:

```ruby
module Ransack
  module Nodes
    class Condition < Node
      # ...
      def casted_array?(predicate)
        predicate.value.is_a?(Array) && predicate.is_a?(Arel::Nodes::Casted)
      end
```

The fix should be pretty simple: just check for the class before calling `#value` on the `predicate`. 

For more context, the previous implementation didn't raise because there was an extra support method that hid the fact that the predicate may not respond to either #val or #value:

```ruby
        def casted_array?(predicate)
          value_from(predicate).is_a?(Array) && predicate.is_a?(Arel::Nodes::Casted)
        end

        def value_from(predicate)
          if predicate.respond_to?(:value)
            predicate.value # Rails 6.1
          elsif predicate.respond_to?(:val)
            predicate.val # Rails 6.0
          end
        end
```

So, in my case this was equivalent to `nil.is_a?(Array) && predicate.is_a?(Arel::Nodes::Casted)`

PR #1468, opened 2023-12-15
Check type before sending `#value` message to predicate

spaghetticode:
Closes https://github.com/activerecord-hackery/ransack/issues/1467

The previous implementation was giving for granted that every predicate respond to `#value`, which doesn't seem to be the case at least when having a `Arel::SelectManager`.

by simply inverting the terms of the existing AND check we can fix the issue without introducing unknown side effects.


spaghetticode:
Interestingly, CI is not running on the PR 🤷

scarroll32:
> Interestingly, CI is not running on the PR 🤷

Needs to be approved for the 1st run.

scarroll32:
@spaghetticode could you please add a test for this change?

spaghetticode:
@scarroll32 I've just added a test that fails without the change. 

The method I modified is private, so it needs to be tested indirectly. I followed the existing examples, but I had to add a new ransacker to the `Person` model, I hope it makes sense. Please let me know if it works for you, thanks!

MadelineCollier:
I am also having this issue, will there be any movement on this PR @scarroll32 ?

```
NoMethodError:
       undefined method `value' for #<Arel::SelectManager:0x000000010b0fb3b8 @ast=#<Arel::Nodes::SelectStatement:0x000000010b0fb390 @cores.........



               predicate.value.is_a?(Array) && predicate.is_a?(Arel::Nodes::Casted)
                        ^^^^^^
     # ./spec/models/spree/product_spec.rb:667:in `block (3 levels) in <top (required)>'

Finished in 28 minutes 11 seconds (files took 4.77 seconds to load)
1 example, 1 failure
```

For now I have locked my gemfile to `ransack 4.1.1` and specs pass on that version.

Issue #1460, opened 2023-11-08
Raising ArgumentError exception on Ransack operations


zokaibr:
Currently, when something goes wrong with some of Ransack operations of Searching and Sorting, it raises an `ArgumentError` exception.

There is a sample on the docs:
```
Article.ransack(unknown_attr_eq: 'Ernie')
# ArgumentError (Invalid search term unknown_attr_eq)
```

I`m proposing here to change it to a custom error specific for Ransack operations because this error is a Ruby internal one specific for missing arguments on a method call.

Besides that, Ransack users will tend to rescue this error to render a response to the end user and it may lead us to a bad error supression.

dfmoreto:
If it's approved, I would like to work on it = )

scarroll32:
Please go ahead @dfmoreto !

moretoend:
@scarroll32 I think it's done. I've opened the PR #1478 with the resolution = )

scarroll32:
Adding to 5.0.0 as it may be a breaking change

scarroll32:
> [!NOTE]
> Comment from **Claude** (Claude Code), acting on behalf of @scarroll32.

Since #1543 (4.3.0) unknown attributes and predicates raise `Ransack::InvalidSearchError`, and 5.0.0 (#1694) does the same for invalid combinators. It subclasses `ArgumentError` so existing rescues keep working, and you can now rescue the Ransack class specifically. Closing as done.

PR #1478, opened 2024-02-14
Set new error to replace ArgumentError

zokaibr:
This PR contains a proposal of removing ArgumentError raising on Search and Sorting and replace it by a custom error Ransack::InvalidSearchError as issued on #1460.

As of now, the exception Ransack::InvalidSearchError is inheriting from ArgumentError to enable a backward compatibility avoiding breaking changes in a minor version and a deprecation message for using ArgumentError were added. In a future version, it is expected to replace this inheritance from ArgumentError to StandardError.

The docs were also updated and there is a proposal of moving Error Handling to Getting Started area as it is an important configuration to be set.

### DIFF
--- a/docs/docs/getting-started/error_handling.md
+++ b/docs/docs/getting-started/error_handling.md
@@ -1,0 +1,73 @@
+---
+sidebar_position: 6
+title: Error Handling
+---
+
+Ransack can be configured to raise errors with attributes, predicates or sorting misusing
+
+## Configuring error raising
+
+By default, Ransack will ignore any unknown predicates or attributes:
+
+```ruby
+Article.ransack(unknown_attr_eq: 'Ernie').result.to_sql
+=> SELECT "articles".* FROM "articles"
+```
+
+Ransack may be configured to raise an error if passed an unknown predicate or
+attributes, by setting the `ignore_unknown_conditions` option to `false` in your
+Ransack initializer file at `config/initializers/ransack.rb`:
+
+```ruby
+Ransack.configure do |c|
+  # Raise errors if a query contains an unknown predicate or attribute.
+  # Default is true (do not raise error on unknown conditions).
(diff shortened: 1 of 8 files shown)


scarroll32:
@moretoend this looks great but it is blocked by a CI problem https://github.com/activerecord-hackery/ransack/issues/1489

scarroll32:
@zokaibr sorry fo the delay on reviewing this. Can you please fix the conflicts and rebase?

scarroll32:
> [!NOTE]

Thanks @zokaibr — closing this because it has already landed, just not through this PR.

Everything substantive here is on `main` today:

- **`lib/ransack/invalid_search_error.rb`** exists, with `InvalidSearchError < ArgumentError` exactly as you wrote it.
- **`Search#build`** raises `InvalidSearchError` instead of a bare `ArgumentError`.
- **`Search#sorts=`** does too.
- **Your specs** are there — `base_spec.rb` has both `raises InvalidSearchError exception if ransack! called with unknown condition` and the `ArgumentError` counterpart, so the subclass relationship stays covered.
- **The documentation** you added ended up in `docs/docs/going-further/other-notes.md`, near-verbatim, including the `ignore_unknown_conditions` example and the `.ransack!` alternative.

Subclassing `ArgumentError` was the right call, and it's why this could ship without a deprecation cycle — existing `rescue ArgumentError` handlers keep working unchanged.

The one part not carried over is the `puts '[DEPRECATED] ...'` lines, and deliberately so: a library writing to stdout on every raise is hard for an application to silence or route, and the subclass makes the warning unnecessary — nothing breaks for anyone rescuing `ArgumentError`. If a hard deprecation is wanted later, `ActiveSupport::Deprecation` is the mechanism for it.

Closing as resolved. Thanks for the work — it made it in.

PR #1485, opened 2024-03-24
Remove ActionView::Helpers::Tags::Base#value Monkey Patch

Aesthetikx:
This library provides an override of `ActionView::Helpers::Tags::Base#value` (the method that is used to determine the current value of a form field built from a template object) that allows calling private methods. The implementation is otherwise the same as the one provide by ActionView.

I ran into some bizarre edge cases with this implementation when using `OpenStruct`, for example just having the `ransack` gem in memory and then using a `f.text_field :test` with an `OpenStruct` as `model` will display a confusing `wrong number of arguments, given 0 expected 2..3` message because `#test` is a private method on `Kernel`. Regardless of how I got here, I don't think this monkey patch is necessary or desired anymore.

Without looking in detail, the reason I believe this patch was created in the first place is because [here](https://github.com/activerecord-hackery/ransack/blob/4a6f9b93adb860c038b6be0c6d5972cb51ea1e70/lib/ransack/helpers/form_builder.rb#L154) we define a select tag with `:p`, and like `:test`, `:p` is a private method on `Kernel`, and creating a select tag in this way will cause the form builder to see if there is a `:p` method to fetch an existing value.

In the second commit in this PR, we can simply default `selected: nil` in `options` such that the form builder does not try to figure out what the existing value is to see if one of the values is selected.

I don't use Ransack directly much (outside of ActiveAdmin) so I can't say if that will be a breaking change, but the tests do pass after that change. Open to feedback!

### DIFF
--- a/lib/ransack/helpers/form_builder.rb
+++ b/lib/ransack/helpers/form_builder.rb
@@ -1,23 +1,5 @@
 require 'action_view'
 
-module ActionView::Helpers::Tags
-  # TODO: Find a better way to solve this issue!
-  # This patch is needed since this Rails commit:
-  # https://github.com/rails/rails/commit/c1a118a
-  class Base
-    private
-    if defined? ::ActiveRecord
-      def value
-        if @allow_method_names_outside_object
-          object.send @method_name if object && object.respond_to?(@method_name, true)
-        else
-          object.send @method_name if object
-        end
-      end
-    end
-  end
-end
-
 RANSACK_FORM_BUILDER = 'RANSACK_FORM_BUILDER'.freeze
 
 require 'simple_form' if
--- a/lib/ransack/helpers/form_builder.rb
+++ b/lib/ransack/helpers/form_builder.rb
@@ -112,6 +112,7 @@ module Ransack
 
       def predicate_select(options = {}, html_options = {})
         options[:compounds] = true if options[:compounds].nil?
+        options[:selected] ||= nil
         default = options.delete(:default) || Constants::CONT
 
         keys =


Aesthetikx:
Adding another comment for clarity, to explain another way:

Ransack is giving form elements short names like `p` for predicates, `c` for conditions, or `g` for groupings. Presumably these end up coming into a Rails controller as params that Ransack can understand. These aren't intended to provide default values to those fields, as e.g. it is unlikely that an object has a `#c` method. That said, ActionView form builder will still see if an object responds to that method (the method matching name of the field) to get a current value, or selected dropdown item. In general nothing happens here because there is no `#c` or `#g`, but there is in fact a `#p` method (the one that prints things, e.g. `p 'Hello World!'`. This method also happens to be a private method, so that you can use it in an IRB or `Main`, but you can't say `Object.new.p 'hello'`. However, you can `Object.new.send(:p)`, which coincidentally does nothing and returns `nil` when no arguments are passed. In the documentation [here](https://apidock.com/rails/ActionView/Helpers/FormOptionsHelper/collection_select) you can see

> The value returned from calling method on the instance object will be selected. If calling method returns nil, no selection is made without including :prompt or :include_blank in the options hash.

and hence because `#p` returns nil, nothing 'bad' happens -- although you do have to allow form builder to call private methods to make the problem 'go away' and presumably hence the reason for the monkey patch in the first place.

scarroll32:
> [!NOTE]

Thanks @Aesthetikx — and your instinct was right on both counts. Rebased as #1690, with you as co-author.

Confirming your diagnosis with a reproduction on current `main`:

```ruby
ActionView::Helpers::FormBuilder.new(:thing, OpenStruct.new, view, {}).text_field(:test)
# ArgumentError: wrong number of arguments (given 0, expected 2..3)
```

With the patch removed it renders normally. So this isn't a strange edge case — Ransack was changing the behaviour of **every form in the host application** just by being loaded, for any attribute name colliding with a private `Kernel` method. #1215 is the same bug reported against `open` and `format`.

Your reasoning about `:p` and `Kernel#p` was also correct, and `options[:selected] ||= nil` is the right way out: it stops the form builder looking up a value it never needed.

> I don't use Ransack directly much (outside of ActiveAdmin) so I can't say if that will be a breaking change

Fair caution. I added a regression spec for the `OpenStruct`/`test` case and checked it fails on `main` and passes with the change, and the full `form_builder_spec` and `form_helper_spec` are green. Removing a monkey patch on a Rails internal still belongs in a major version, so #1690 is aimed at 5.0.0 rather than the 4.5.0 release going out now.

Closing in favour of #1690. Thanks for the careful write-up — the `TODO: Find a better way to solve this issue!` in that patch had been sitting there a long time.

PR #1501, opened 2024-06-29
Add support for Trilogy database adapter

### LINKED ISSUE #1500: Support for Trilogy database adapter
When wildcard matching, ransack takes care to escape any special characters in the search term.  This is from `lib/ransack/constants.rb`:
```
module_function
  # replace % \ to \% \\
  def escape_wildcards(unescaped)
    case ActiveRecord::Base.connection.adapter_name
    when "Mysql2".freeze
      # Necessary for MySQL
      unescaped.to_s.gsub(/([\\%_])/, '\\\\\\1')
    when "PostgreSQL".freeze
      # Necessary for PostgreSQL
      unescaped.to_s.gsub(/([\\%_.])/, '\\\\\\1')
    else
      unescaped
    end
  end
end
```
There is a new MySQL adapter called Trilogy: https://github.blog/2022-08-25-introducing-trilogy-a-new-database-adapter-for-ruby-on-rails/

Can we get support for this adapter?  I'm not sure how much of the ransack code would be impacted, this (wildcard search escaping) just happened to be the thing I ran into.

navels:
Fixes #1500

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Add support for Trilogy database adapter by @navels in https://github.com/activerecord-hackery/ransack/pull/1501
+
 ## 4.1.0 - 2023-10-23
 
 ### 🚀 Features
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,17 +73,18 @@ Here's a quick guide:
    bundle exec rake spec
    ```
 
-   The test suite runs by default with SQLite3. To run the test suite with PostgreSQL or MySQL, use:
+   The test suite runs by default with SQLite3. To run the test suite with PostgreSQL, MySQL (Mysql2 adapter), or MySQL (Trilogy adapter), use:
 
    ```sh
    DB=pg bundle exec rake spec
    DB=mysql bundle exec rake spec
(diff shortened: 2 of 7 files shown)


scarroll32:
> [!NOTE]

Thanks @navels — correct diagnosis. Trilogy is MySQL, but the checks matched the literal `"Mysql2"`, so Trilogy users fell through to the non-MySQL branch and got **no wildcard escaping at all**. That's the same silent wrong-results class as #1581. Rebased as #1698, with you as co-author.

The one thing I added is a way to know it works. #1501 widened the adapter checks but there was no Trilogy connection in the spec schema and no CI job, so "supports Trilogy" would have stayed an assertion — and since the original failure mode was silent, a regression would go unnoticed the same way the bug did.

#1698 adds a `DB=trilogy` connection, the `trilogy` gem, a CI job running the full suite on Rails 8.0 and 8.1, and the `CONTRIBUTING.md` line. I don't have a MySQL server here, so that job passing is the actual evidence — if it fails, it shouldn't merge.

Worth noting: #1682 removes the adapter switch in `escape_wildcards` altogether, escaping uniformly and emitting an explicit `ESCAPE` clause. If that lands first it fixes the Trilogy wildcard bug for free, for every adapter at once. The connection, spec helpers and CI job are still needed.

Closing in favour of #1698.

PR #1502, opened 2024-07-08
Deep stripping whitespace in nested search params with group conditions

### LINKED ISSUE #1414: strip_whitespace doesn't work with group conditions
```
User.ransack({g: [name_cont: "  John  ", email_cont: "  John ", m: "or"]}).result
```

`"  John "` should be strip whitespace but it doesn't

aukraine:
Resolves #1414 

Fix case when `strip_whitespace` doesn't work with group conditions

**Before**
```
> User.ransack({g: [name_cont: "  John  ", email_cont: "  John ", m: "or"]}).result
=> []
```

**After**
```
> User.ransack({g: [name_cont: "  John  ", email_cont: "  John ", m: "or"]}).result
=> [#<User id, ...>]
```


scarroll32:
> [!NOTE]

Thanks @aukraine, and sorry this sat for over a year — it's a one-word fix for a real inconsistency, and `deep_transform_values` is exactly the right call.

Rebased onto current `main` as #1688, with you as co-author. The only change: your branch conflicted with #1657, which added `&& !i.nil?` to the `delete_if` on the adjacent line. Taking your side of the conflict wholesale would have reverted that and regressed `name_in: [nil]`, so both are kept.

Closing in favour of #1688.

PR #1506, opened 2024-07-15
Fix require of rails 7.2 join association file

robingram:
Don't require the file for rails > 7.1.0 but < 7.2.

Fixes #1505.

PR #1507, opened 2024-07-21
fix wrong require

42thcoder:
If the ActiveRecord version is 7.1.3.4, it will lead to the loading of the join_association_7_2 module, which will then cause a similar NoMethodError: undefined method 'with_connection' for class Ignore error.


robingram:
Duplicates #1506

scarroll32:
Closing as dup

PR #1511, opened 2024-07-28
Add a note for macOS users to Contributing to Ransack to make it more friendly.

satou-haruka-37:
As per Contributing to Ransack
I forked the repository and ran `bundle install`,
I got the following error
````
1 warning generated.
compiling statement.c
linking shared-object mysql2/mysql2.bundle
ld: warning: ignoring duplicate libraries: '-lruby.3.2'.
'-lruby.3.2'.
ld: library 'zstd' not found
clang: error: linker command failed with exit code
1 (use -v to see invocation)
make: *** [mysql2.bundle] Error 1

make failed, exit code 2
````

Therefore, I thought it would be better to add an annotation.


kou:
It seems that the error message is reported for mysql2 gem not ransack gem.
Is https://github.com/brianmario/mysql2/issues/1345 the same problem?

scarroll32:
Should this be closed [satou-haruka-37](https://github.com/satou-haruka-37) ?

satou-haruka-37:
Sorry for the late reply.
I am trying to investigate and learn about the cause of mysql and progress has been halted.
If mysql2 is the root cause, and there was no need to include a description on the ransack, then this pull request can be closed. My apologies.

PR #1514, opened 2024-08-05
Make polyamorous/activerecord/join_association including work properly with Rails 7.1 builds

velimovich:
Include code intended for Rails 7.2 only when actual Rails version is 7.2+

Currently it requires this file on Rails `7.1.1+` for example


velimovich:
Close in favour of https://github.com/activerecord-hackery/ransack/pull/1506

PR #1523, opened 2024-09-10
fix ActiveModel::RangeError when using multiple search fields

kha-wogi:
`ActiveModel::RangeError` is currently raised when performing searches with large integer values across multiple fields


scarroll32:
> [!NOTE]

Thanks @kha-wogi — correct diagnosis and a good spec. Rebased as #1691, with you as co-author.

For the record, since the "why" isn't obvious from the diff: the unwrapping ran once *after* the per-attribute predicates were reduced. One attribute meant the reduced node was the predicate itself, so it worked; several meant `reduce(:or)` produced an `Or` whose `right` is another node rather than a `Casted`, so `replace_right_node?` returned false and nothing happened. Moving it inside the `map` puts the check where the `Casted` actually is.

One change during the rebase: your diff restructured the whole method, and `main` has since grown the `not_nested_condition` guard and the `not_cont` branch of that `case`. Taking your version wholesale would have reverted those, so I reapplied just the structural part — hoisting the result into a local and moving the range check into the block — onto the current body.

Your spec is kept as-is. I checked it raises `ActiveModel::RangeError` against `main` and passes with the fix.

Closing in favour of #1691.

PR #1512, opened 2024-08-01
Fix `ransack_alias` issue, close #1239

itsalongstory:

Issue #916, opened 2018-05-03
search_field don't preserve values between requests for alias


fidalgo:
Hi!
First of all thanks for your work :tada: 

I'm using the ransack (1.8.8) with a Rails 5.1 and Ruby 2.4.3 with the following code:

Ransack alias:
`ransack_alias :quick_search, :sequence_or_order_customer_name_or_order_customer_vat_identification_or_order_line_items_description`

and this form:
```
        <%= f.search_field :quick_search_cont, value: params.dig(:q, :quick_search_cont), class: "form-control"%>
        <span class="input-group-btn">
          <%= button_tag type: :search, class: 'btn btn-primary' do %>
            <%= t('actions.search') %>
            <span class="fas fa-search"></span>
          <% end %>
        </span>
......
    <% end %>
```

and I need to add `value: params.dig(:q, :FIELD_NAME)` to every `search_field` in my form, to have the values binded in the form, ie, when the results are returned and the page rendered I have the search input value filled.

Where is an url example:
`http://localhost:3000/invoices?utf8=%E2%9C%93&q%5Bquick_search_cont%5D=&button=&q%5Binvoice_search_cont%5D=&q%5Bperson_cont%5D=dias&q%5Bproduct_search_cont%5D=`

Am I doing something wrong or there is a bug in the gem?

supairish:
I'm seeing this also

gregmolnar:
@fidalgo @supairish Do you also see this on 2.1.1?

supairish:
@gregmolnar I'm not sure, i'm linking to master in our Gemfile.  I'll see if i have time today to try 2.1.1

jean-francois-labbe:
I see this too with `2.1.1`

teoulas:
2.3.0 also has the same bug:

```ruby
class SomeModel
  ransack_alias :abc, :name_or_address_or_something_else
end

r = SomeModel.ransack({abc_cont: "xyz"})
r.abc_cont
# => nil
r.name_or_address_or_something_else_cont
# => "xyz"
```

mike-northsail:
This is still an issue on latest master

jonsgreen:
I am curious if there is anyway that I can assist in resolving this bug?

scarroll32:
> [!NOTE]
> Comment from **Claude** (Claude Code), acting on behalf of @scarroll32.

[...]

Issue #689, opened 2016-06-23
Ransack form search field value not persisted between requests when using alias


fidalgo:
When I search, using an alias I cannot get the value of the searched parameter in the field.
In the name field I have tested and it's working fine.
My code:
In controller:

```
  def index
    @q = Customer.ransack(params[:q])
    @customers = @q.result(distinct: true)
  end
```

In model:

```
  ransack_alias :quick_search, :name_or_phone_or_mobile_or_email_or_vat_identification
```

In view:

```
      <%= search_form_for @q do |f| %>
      <%= f.search_field :quick_search_cont, class: "form-control"%>
    </div>
    <div id="table-filter" class="well filter-well">
      <div class="form-group">
        <%= f.label (:name) %>
        <%= f.search_field :name_cont, class: "form-control" %>
      </div>
```

Ransack version:
GIT
  remote: git://github.com/activerecord-hackery/ransack.git
  revision: 64d8c402e6e79e8ec2966a15c35b98196813ef8e

I'm using Ransack directly with:
Database: PostgreSQL
Rails: 5.0.0.rc1
ruby 2.2.5

moofkit:
same here

Ransack version:
GIT
remote: git://github.com/activerecord-hackery/ransack.git
revision: fd94685abc0d18d30a730a57fb3dbd8aff048f7d

Database: PostgreSQL
Rails: 5.0.0
ruby 2.3.1

luislezcair:
Any update on this? I'm having the same problem.

doubleton:
If it helps anyone, I worked around this using `value` option:
```
= f.search_field :quick_search_cont, value: params.dig(:q, :quick_search_cont)
```

asaletnik:
Same here.

Ransack: 1.8.3 / 1.8.2
Database: MySQL
Rails: 4.2.9
Ruby: 2.3.3

mattwbarry:
Same here

Ransack: 1.8.3
Database: PostgreSQL
Rails: 4.0.13
Ruby: 2.3.3

bitberryru:
Same bug here

Ransack: 2.1.1
Database: MySQL
Rails: 5.2.2
Ruby: 2.5.3

Also ransack sorting ignores aliases.

danielnolan:
This is happening for me also when using ransack with simple form, using an alias doesn't populate my checkboxes when the page refreshes after submitting the form. If I remove the alias everything works as expected
Ransack: 3.1.0
[...]

PR #1525, opened 2024-09-13
Correct ransack_alias to work with form

le0pard:
Issue itself: https://github.com/activerecord-hackery/ransack/issues/916

also should fix https://github.com/activerecord-hackery/ransack/pull/1512 (no need compare values)


lucasbiguet:
any update on this ? it’s currently very confusing to work with rails `form_for` and `RansackSearch` since the value cannot get detected properly by the form helpers due to `respond_to_missing?` not being implemented.

scarroll32:
> [!NOTE]

Thanks @le0pard — closing this as superseded, with the reasoning so you can push back if I've missed a case.

The bug this addresses (an aliased field losing its value between requests — #689, #916) was fixed today by #1693, which took the narrower route from #1529: `Grouping#read_attribute` resolves the alias when the name it's given holds nothing. Verified end-to-end through a real `search_form_for`: `f.text_field :daddy_cont` now renders `value="John"`. That covers the shorthand alias, and because it strips the predicate and re-resolves against the full aliased attribute string, it also covers aliases onto multi-attribute targets like `:author_first_name_or_author_last_name`.

Your approach — remembering the original name on the condition via an `n:` field, threaded through `extract`/`build`/`eql?`/`hash` and `Grouping#[]=` — is more principled, but it's an 11-file change to the condition model for a problem the 7-line fix now handles, and after today's merges it conflicts in `grouping.rb`. If you have a form scenario that's still broken on current `main`, a failing spec would be very welcome and I'll look again.

PR #1526, opened 2024-09-14
Fix inverts search, when have additional conditions in scope (like default_scope)

le0pard:
Fix code for test, which should pass


le0pard:
looks like github action have issue with report, because all tests passed (no 9 expected checks)

le0pard:
https://github.com/le0pard/ransack/pull/3 - tests green

scarroll32:
> [!NOTE]

Thanks @le0pard — closing this as superseded, with the checking shown so you can disagree if I've got it wrong.

The spec this PR set out to make pass, `preserves (inverts) default scope and conditions for negative subqueries`, **already passes on `main`**. Your diff's only change to it is deleting the comment above it:

```diff
-      # The failure/oversight in Ransack::Nodes::Condition#arel_predicate or deeper is beyond my understanding of the structures
```

It was fixed by later work on negative subqueries — `not_nested_condition` and the `not_cont` branch in `Condition#arel_predicate`, plus the polymorphic/default_scope handling in #1561. Verified on current `main`:

```
$ bundle exec rspec spec/ransack/search_spec.rb -e "preserves"
2 examples, 0 failures
```

With that already resolved, the remaining effect of this PR is the `subquery.where(join_conditions)` line, and that **breaks two specs that `main` has since gained**:

```
1) Context#build_correlated_subquery handles Arel::Nodes::And with children
   expected: 1
        got: 2
2) Context#build_correlated_subquery correctly extracts correlated key from complex AND conditions
   expected: 1
        got: 2
```

Those assert that the correlated subquery carries exactly one constraint; re-adding the join conditions gives it two.

So the problem is fixed and this approach now conflicts with how it was fixed. Closing on that basis — but if you have a case that still misbehaves on current `main`, please open an issue with it and I'll look again. Your diagnosis of the underlying problem was right, and the comment you deleted was fair at the time.

PR #1528, opened 2024-10-01
Fix Japanese Locales

ryoya1122:
fix japanese locales

The current translations are all accompanied by a conjunction, such as “is equals”, which is not an appropriate choice for the select box, so this has been corrected.


scarroll32:
Thank you @ryoya1122

Issue #1520, opened 2024-08-29
ransack_alias doesn't work in associations


matgaw:
class Parent
      ransack_alias :full_name, :first_name_or_last_name
    end
    
    class Child
      belongs_to :parent
    end

Working scenario:

`Parent.ransack(full_name_cont: 'a').result # works correctly`

Error scenario:
`Child.ransack(parent_full_name_cont: 'a').result`
Result:

    undefined method `type' for nil:NilClass (NoMethodError)

          attr.klass.columns.find { |column| column.name == name }.type
                                                                  ^^^^^

dougjq:
This particular line of code is problematic for ActiveAdmin users, as well. If you use ActiveAdmin #includes on a has_one through, you will trigger a call to ransack. ActiveAdmin is feeding ransack the association through the parent, with the result that ransack is told to expect to find grandparent_id that does not exist. This line throws undefined method `type' for nil:NilClass.

Suppose class Whatever < ApplicationRecord has_one :grandparent, through: :parent. ActiveAdmin sends ransack instructions to look for whatevers.grandparent_id. 

I can't help but feel there is inadequate error checking here in ransack. Further up the trace in

lib/ransack/nodes/condition.rb:25:in `extract' 

there is this comment:

            # TODO: Figure out what to do with multiple types of attributes,
            # if anything. Tempted to go with "garbage in, garbage out" here.

Not being a ransack expert I can't say what the solution would be. But I can say this is a new error introduced since rack 3 and it seems improperly handled.

8bit-mate:
I'd like to point out that "mixed" compound conditions that contain an alias + an association do not work, too.

Consider the following setup:

```ruby
class Author
  has_many :books, dependent: :destroy

  def self.ransackable_attributes(auth_object = nil)
    %w[full_name]
  end
end

class Book < ApplicationRecord
  ransack_alias :ttl, :full_title

  belongs_to :author

  def self.ransackable_associations(auth_object = nil)
    %w[author]
  end

  def self.ransackable_attributes(auth_object = nil)
    %w[full_title ttl]
  end
end
```

Now, with this setup we have:

```erb
<%= search_form_for @q do |f| %>
  <%# ✅ Case #1: alias for the model's attribute - works %>
  <%= f.search_field :ttl_cont, placeholder: "Filter by full_title..." %>

  <%# ✅ Case #2: association's attribute - works %>
[...]

PR #1529, opened 2024-10-03
Add support for Ransack::Nodes::Grouping#read_attribute to use aliases

### LINKED ISSUE #689: Ransack form search field value not persisted between requests when using alias

DougEdey-Filescom:
Fixes https://github.com/activerecord-hackery/ransack/issues/689

This works locally in my test environment, it should be lightweight and only used if we don't have an immediately matching attribute in the context.

It's a pretty simple change but I'm not sure if there's some unforeseen consequences I'm unaware of.

I looked at the contributing guide but there's nothing about updating the changelog in the PR :)

### DIFF
--- a/lib/ransack/nodes/grouping.rb
+++ b/lib/ransack/nodes/grouping.rb
@@ -178,6 +178,13 @@ module Ransack
       end
 
       def read_attribute(name)
+        if self[name].nil?
+          stripped_name = name.dup
+          predicate = Predicate.detect_and_strip_from_string!(stripped_name)
+          aliased_attribute = context.ransackable_alias(stripped_name)
+          name = "#{aliased_attribute}_#{predicate}" unless aliased_attribute == stripped_name
+        end
+
         if self[name].respond_to?(:value)
           self[name].value
         else
--- a/spec/ransack/nodes/grouping_spec.rb
+++ b/spec/ransack/nodes/grouping_spec.rb
@@ -106,6 +106,23 @@ module Ransack
         end
       end
 
+      describe "#read_attribute" do
+          let(:conditions) do
+            {
+              '0' => {
+                'a' => { '0'=> { 'name' => 'parent_name', 'ransacker_args' => '' } },
+                'p' => 'cont',
+                'v' => { '0' => { 'value' => 'John' } }
+              },
+            }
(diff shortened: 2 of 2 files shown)


scarroll32:
> [!NOTE]

Thanks @DougEdey-Filescom — rebased as #1693, with you as co-author. Sorry it sat so long for a seven-line fix to a bug open since 2016.

> I'm not sure if there's some unforeseen consequences I'm unaware of

I looked for them and didn't find any. The property that makes it safe is the one you identified — the fallback only runs when `self[name]` is nil, so the common path does no extra work and existing behaviour is untouched. The full suite is green, including `form_helper_spec` and `form_builder_spec`, which is where a regression here would surface.

I verified your spec fails against `main` (returns `nil` instead of `"John"`) and passes with the change, and tidied its indentation.

> I looked at the contributing guide but there's nothing about updating the changelog in the PR :)

You were right — and still are. Since 4.4.0 the changelog is generated from GitHub Release notes, so PRs don't touch `CHANGELOG.md` at all.

Worth noting this is the same underlying problem as #916 and #1520. #1525 also targets aliases in forms with a broader change across 11 files; landing your narrower fix first is the safer order.

Closing in favour of #1693.

PR #1534, opened 2024-10-29
Fix rails 7.2 loading

### LINKED ISSUE #1518: Rails 7.2 - LoadError: cannot load such file -- polyamorous/activerecord_7.2_ruby_2/join_association (LoadError)
Upgrade to rails 7.2 and you will get the following message whilst constructing the database...

bin/rails aborted!
LoadError: cannot load such file -- polyamorous/activerecord_7.2_ruby_2/join_association (LoadError)
Did you mean?  polyamorous/activerecord_7.0_ruby_2/join_association
               polyamorous/activerecord_6.1_ruby_2/join_association
               polyamorous/activerecord_6.0_ruby_2/join_association

jdlubrano:
Fixes #1518 

# Problem

In a typical Rails app, the Ransack gem gets loaded before ActiveRecord gets loaded.  [Rails recommends](https://guides.rubyonrails.org/configuring.html#avoid-loading-rails-frameworks) that gems avoid loading Rails frameworks when they themselves are loaded.  Ransack accomplishes that (as far as I can tell); however, Ransack initializes certain parts of its code within blocks like

```ruby
if defined?(::ActiveRecord)
  ...set up some stuff
end
```

If Ransack is loaded before `ActiveRecord` is loaded, those blocks of code are never reached.  I think these blocks of code should be wrapped with `ActiveSupport.on_load(:active_record)` instead of `if defined?(::ActiveRecord)`.  The blocks of code should run after `ActiveRecord` is loaded, which is what `on_load(:active_record)` is meant to accomplish.

# Test coverage

I wasn't quite sure how to cover this in the specs, so I adjusted the load order in `spec/spec_helper.rb` to `require "ransack"` first.  That did expose the load order issue(s) within the gem, so maybe that is good enough.  It might be ideal to set up a test that spins up a Rails app and runs a basic query just to check that the basic functionality of the gem works within Rails 🤷‍♂️ .  For local development, I used a single file Rails app like this:

```ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  gem 'debug'
  gem 'irb'
  gem 'rails', '>= 7.1'
  gem 'sqlite3', '~> 1.4'
  gem 'ransack' # , path: "./ransack"
end

require 'debug'
require 'rails/all'
database = 'development.sqlite3'

ENV['DATABASE_URL'] = "sqlite3:#{database}"

class App < Rails::Application
  config.load_defaults 7.2
  config.eager_load = false
end

ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: database)

  create_table :posts, force: true do |t|
    t.string :title, null: false
  end
end

App.initialize!

  def self.ransackable_associations(_auth_object = nil) = []

  def self.ransackable_attributes(_auth_object = nil)
    %w[title]
  end

end

1.upto(10) { |i| Post.create!(title: "Title #{i}") }

IRB.start

Post.ransack(title_eq: "Title 1").result.to_sql
=> SELECT * FROM posts where title = 'Title 1'
```

### DIFF
--- a/lib/ransack.rb
+++ b/lib/ransack.rb
@@ -1,3 +1,4 @@
+require 'active_support/dependencies/autoload'
 require 'active_support/core_ext'
 require 'ransack/configuration'
 require 'polyamorous/polyamorous'
--- a/lib/ransack.rb
+++ b/lib/ransack.rb
@@ -1,4 +1,11 @@
 require 'active_support/dependencies/autoload'
+require 'active_support/deprecation'
+require 'active_support/version'
+
+if ::ActiveSupport.version >= ::Gem::Version.new("7.1")
+  require 'active_support/deprecator'
+end
+
 require 'active_support/core_ext'
 require 'ransack/configuration'
 require 'polyamorous/polyamorous'
--- a/lib/ransack/helpers/form_builder.rb
+++ b/lib/ransack/helpers/form_builder.rb
@@ -6,13 +6,12 @@ module ActionView::Helpers::Tags
   # https://github.com/rails/rails/commit/c1a118a
   class Base
     private
-    if defined? ::ActiveRecord
-      def value
-        if @allow_method_names_outside_object
-          object.send @method_name if object && object.respond_to?(@method_name, true)
-        else
-          object.send @method_name if object
-        end
+
+    def value
+      if @allow_method_names_outside_object
+        object.send @method_name if object && object.respond_to?(@method_name, true)
+      else
+        object.send @method_name if object
       end
     end
   end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,13 +1,12 @@
-require 'machinist/active_record'
-require 'polyamorous/polyamorous'
+require 'ransack'
 require 'sham'
 require 'faker'
-require 'ransack'
 require 'action_controller'
 require 'ransack/helpers'
 require 'pry'
 require 'simplecov'
 require 'byebug'
+require 'machinist/active_record'
 
 SimpleCov.start
 I18n.enforce_available_locales = false
--- a/lib/polyamorous/polyamorous.rb
+++ b/lib/polyamorous/polyamorous.rb
@@ -1,4 +1,4 @@
-if defined?(::ActiveRecord)
+ActiveSupport.on_load(:active_record) do
   module Polyamorous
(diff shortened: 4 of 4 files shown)


lib/ransack.rb
```diff
@@ -1,3 +1,4 @@
+require 'active_support/dependencies/autoload'
```
jdlubrano, line 1: Just requiring this explicitly since `on_load` does get used further down in this file (and I suppose should be considered in any other files that get required from here).

lib/ransack/helpers/form_builder.rb
```diff
@@ -6,13 +6,12 @@ module ActionView::Helpers::Tags
   # https://github.com/rails/rails/commit/c1a118a
   class Base
     private
-    if defined? ::ActiveRecord
```
jdlubrano, line 9: I'm not sure that this was ever truly dependent on ActiveRecord.  At one point, this was a version check.  I agree with the comment that it would be nice to avoid this patch to ActionView, but I also think that the patch is necessary regardless of whether ActiveRecord is defined or not.  Moreover, the code within doesn't depend on ActiveRecord as we are just calling `send` on an object 🤔 .

spec/spec_helper.rb
```diff
@@ -1,13 +1,12 @@
...
+require 'ransack'
 require 'sham'
 require 'faker'
-require 'ransack'
 require 'action_controller'
 require 'ransack/helpers'
 require 'pry'
 require 'simplecov'
 require 'byebug'
+require 'machinist/active_record'
```
jdlubrano, line 9: `machinist/active_record` requires `active_record`, which made all of the code wrapped in `defined?(::ActiveRecord)` execute.  Requiring `machinist/active_record` after required `ransack`, I think, defines a load order that better imitates a Rails app.

jdlubrano:
Hello 👋 .  I was just checking whether there was any interest in this pull request?  I was able to get all of the different DBs and Rails versions passing locally.  Sorry for those oversights initially.

gregmolnar:
@jdlubrano Could you please rebase against main?

gregmolnar:
Thank you!

danhnguyenvh:
@gregmolnar 
Thank you for quick fixing.
Could i ask: do you have plan for the release of this fixing?

PR #1539, opened 2024-11-21
Fix issue with ransackable_attribute not handling symbols correctly

### LINKED ISSUE #1538: Bug?: ransackable_attributes Fails with Symbol Array
When using the ransackable_attributes method in ActiveRecord, passing an array of symbols causes the search conditions for the corresponding attributes to be ignored. However, passing an array of strings works as expected.

### Steps to Reproduce
1: Create a migration for a users table with a name column:
```rb
class CreateUsers < ActiveRecord::Migration[7.1]
  def change
    create_table :users do |t|
      t.string :name

      t.timestamps
    end
  end
end
```

2: Define the User model with ransackable_attributes returning a symbol array:

```rb
class User < ApplicationRecord
  def self.ransackable_attributes(auth_object = nil)
    [:name]
  end
end
```

3: Attempt to perform a Ransack query with a condition on the name attribute:

`User.ransack(name_cont: "hoge").result.to_sql`

## Actual Behavior
The generated SQL query ignores the condition for name:

`SELECT "users".* FROM "users"`

I expected `SELECT "users".* FROM "users" WHERE "users"."name" LIKE '%hoge%'`

===

The code below returned the expected results:

```rb
class User < ApplicationRecord
  def self.ransackable_attributes(auth_object = nil)
    ["name"]
  end
end
```

===
Ruby version: 3.3.6
Rails version: 7.2.2
ransack version: 4.2.1

tatematsu-k:
Fix: https://github.com/activerecord-hackery/ransack/issues/1538

If the behavior is as intended, please close this PR.

---

## Problem
When using the ransackable_attributes method in ActiveRecord, passing an array of symbols causes the search conditions for the corresponding attributes to be ignored. However, passing an array of strings works as expected.

### Steps to Reproduce
1: Create a migration for a users table with a name column:
```rb
class CreateUsers < ActiveRecord::Migration[7.1]
  def change
    create_table :users do |t|
      t.string :name

      t.timestamps
    end
  end
end
```

2: Define the User model with ransackable_attributes returning a symbol array:

```rb
class User < ApplicationRecord
  def self.ransackable_attributes(auth_object = nil)
    [:name]
  end
end
```

3: Attempt to perform a Ransack query with a condition on the name attribute:

`User.ransack(name_cont: "hoge").result.to_sql`

## Actual Behavior
The generated SQL query ignores the condition for name:

`SELECT "users".* FROM "users"`


tatematsu-k:
This PR has been inactive for a while.
Is there anything I can do to help move it forward?

ypoloni:
@tatematsu-k  Thanks a lot! That works well for the 'where' search. However, I also need to override Nodes::Sort with the same solution to get Ransack sorting working as well.

Could you update your pull request changing Nodes::sort too?

```
Ransack::Nodes::Sort.class_eval do
  def valid?
    bound? && attr &&  context.ransackable_attribute?(attr_name.split('.').last, context.klassify(parent))
  end
end
```

tatematsu-k:
@ypoloni 
Thanks comments!
I’ve applied the requested changes.
Please let me know if anything else needs to be updated.

scarroll32:
> [!NOTE]

Thanks @tatematsu-k, and sorry this sat for so long. The behaviour is not intended — `ransackable_attributes` returning symbols should work exactly like strings, and #1538 is a real bug.

The fix is right: `Attribute#valid?` was doing a raw `Array#include?` on the allowlist, so `[:name].include?("name")` was false. Routing it through `Context#ransackable_attribute?`, which already normalises with `to_sym`, makes the two forms equivalent.

Splitting `ransortable_attribute?` out of `ransackable_attribute?` and calling both where sorting is in play is the right call too — it keeps search restricted to `ransackable_attributes` while sorting still consults `ransortable_attributes`, rather than the previous conflation.

Verified locally merged into current `main` — Rails 8.1.3 / Ruby 3.4.9 / SQLite: **527 examples, 0 failures, 1 pending**. CI is green across all 25 checks.

PR #1545, opened 2025-02-06
Rails 7.2.2.1 doesn't need 7.2 join association

gregmolnar:


guigs:
gemspec still have this requirements as dependencies:

  s.add_dependency 'activerecord', '>= 6.1.5'
  s.add_dependency 'activesupport', '>= 6.1.5'

Should this also be incremented? What exactly Rails support is being dropped?

gregmolnar:
I will update that(or you can open a PR for it if you'd like), thanks for spotting it. The latest version doesn't support anything below Rails  7.1

guigs:
Thanks, but even if we fix that for next version, bundler will still happily install ransack 4.3.0 along with Rails 7.0. What kind of issues can we expect?

guigs:
Our build and all tests are passing with 4.3.0 and Rails 7.0.

gregmolnar:
Some of the 7 versions will fail, or at least they fell on CI here and I didn't want to fix the issues since those versions of Rails are not supported anymore anyways.

PR #1559, opened 2025-05-19
Fix enum support

### LINKED ISSUE #1555: Why Doesn’t Ransack Support Rails Enums Properly?
Is this the intended behavior?
```ruby
class Lesson < ApplicationRecord
  enum :status, {
    one: 1,
    two: 2,
    three: 3
  }
  class << self
      ["status"]
    end
  end
end
```

```ruby
Lesson.ransack({ "status_eq" => "two" }).result 
# I expected: SELECT * FROM lessons WHERE status = 2
```
But the actual query is:
```sql
SELECT "lessons".* FROM "lessons" WHERE "lessons"."status" = 0
```

I understand that defining a `ransacker` in the model can be used as a workaround.  
That said, if this is considered a bug, I would truly appreciate any fix or guidance.  
If there's any way I could contribute to resolving it, I’d be more than happy to help.

I’m not very fluent in English, so I apologize if my words sounded rude

matheussilvasantos:
Closes #1555.

### DIFF
--- a/lib/ransack/nodes/condition.rb
+++ b/lib/ransack/nodes/condition.rb
@@ -226,7 +226,7 @@ module Ransack
       end
 
       def casted_values_for_attribute(attr)
-        validated_values.map { |v| v.cast(predicate.type || attr.type) }
+        validated_values.map(&:cast_array)
       end
 
       def formatted_values_for_attribute(attr)
--- a/lib/ransack/nodes/value.rb
+++ b/lib/ransack/nodes/value.rb
@@ -43,6 +43,14 @@ module Ransack
         end
       end
 
+      def cast_array
+        if value.is_a?(Array)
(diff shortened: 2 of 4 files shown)


deivid-rodriguez:
Thank for you getting this started @matheussilvasantos, I think I figured it out and made your spec pass.

pekopekopekopayo:
GOOD 👍  Thank You!

deivid-rodriguez:
@matheussilvasantos Awesome, thanks for trying it out. I think your attribution should still be kept because having a failing spec is an important part of getting something fixed!

deivid-rodriguez:
Also added @pekopekopekopayo's attribution for the spec improvements 👍.

deivid-rodriguez:
Let's do this! I'll try find some time to release it.

PR #1565, opened 2025-06-03
Cast array input to integer node value as nil

### LINKED ISSUE #1556: NoMethodError undefined method 'to_i' for an instance of Array
The error happens at https://github.com/activerecord-hackery/ransack/blob/9c66f49fc1be1a570d7b5ff1768b0bbc4a4f4524/lib/ransack/nodes/value.rb#L83

The error happens when passing an array instead of a string/int to an `_eq` predicate.

The implementation of this Value class doesn't handle errors as it should. It rescues things like `Date.new(y, m, d) rescue nil`, yielding invalid search results eventually.

I would say it has to do exactly the opposite, all those `rescue nil` should be gone, raising errors when invalid data is submitted. Those errors have to be handled in the application, not in the library.

A new error class can be introduced, e.g., `Ransack::InvalidPropertyError`. This error can be raised when the submitted property doesn't match the expectations.
```ruby
def cast_to_integer(val)
  raise Ransack::InvalidPropertyError, "..." unless val.respond_to?(:to_i)
  val.blank? ? nil : val.to_i
end
```

Date parsing should rely on RFC formats and raise otherwise, rather than trying some magick tricks, rescuing, and silently swallowing the errors, yielding invalid search results.

jdufresne:
Ensure malformed or malicious user input doesn't crash the Rails page.

Previously, when a user submits a non-empty array, it would raise the exception:

```
NoMethodError: undefined method 'to_i' for an instance of Array
```

Fixes #1556

### DIFF
--- a/lib/ransack/nodes/value.rb
+++ b/lib/ransack/nodes/value.rb
@@ -88,7 +88,7 @@ module Ransack
       end
 
       def cast_to_integer(val)
-        val.blank? ? nil : val.to_i
+        val.respond_to?(:to_i) && !val.blank? ? val.to_i : nil
       end
 
       def cast_to_float(val)
--- a/spec/ransack/nodes/value_spec.rb
+++ b/spec/ransack/nodes/value_spec.rb
@@ -71,6 +71,18 @@ module Ransack
         end
       end
 
+      [[], ["12"], ["101.5"]].each do |value|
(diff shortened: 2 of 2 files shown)


scarroll32:
Thanks for your contribution!

PR #1570, opened 2025-06-30
Respect `config.active_record.permanent_connection_checkout`

joshuay03:
Context:
- https://github.com/rails/rails/pull/50793
- https://github.com/rails/rails/pull/51349
- https://guides.rubyonrails.org/configuring.html#config-active-record-permanent-connection-checkout
- https://github.com/rails/rails/blob/1b4484c43f32a42824d3de32e22029207f76bc42/activerecord/lib/active_record/connection_handling.rb#L273-L295
- https://github.com/rails/rails/pull/51083

This PR addresses two things:
1. `ActiveRecord::Base.connection` has been soft deprecated in favour of `.lease_connection`, and applications can explicitly enable the deprecation by setting `config.active_record.permanent_connection_checkout` to `:deprecated`, or disallow its usage altogether by setting it to `:disallowed`. In such cases the gem should not be using the deprecated method, but it would be good to avoid using it altogether so no further changes will be needed when it's hard deprecated.
2. Applications can choose to not use permanent connection checkouts altogether (i.e. by not using `.lease_connection`) in favour of the `ActiveRecord::ConnectionAdapters::ConnectionPool#with_connection` and newly added `ActiveRecord::ConnectionHandling#with_connection` APIs. I think it would be good to support such applications by utilising these APIs where possible, instead of leasing a connection. I've ensured that this is true for all non-test code, but I don't know of a good way to prevent further usage (see [this relevant comment](https://github.com/rails/rails/pull/51349#issuecomment-3010939543)). Also note that `ActiveRecord::ConnectionHandling#with_connection` was added in 7.2, and since this gem currently supports 7.1 I've only relied on the connection pool method.


bug_report_templates/test-ransack-scope-and-column-same-name.rb
```diff
@@ -39,7 +39,7 @@
 # Display versions.
 message = "Running test case with Ruby #{RUBY_VERSION}, Active Record #{
   ::ActiveRecord::VERSION::STRING}, Arel #{Arel::VERSION} and #{
-  ::ActiveRecord::Base.connection.adapter_name}"
+  ::ActiveRecord::Base.adapter_class::ADAPTER_NAME}"
```
joshuay03, line 42: I've opted to use this constant even where we could lease a connection (i.e., here and tests) for consistency.

joshuay03:
Bump 🙏🏽

scarroll32:
> [!NOTE]

Thanks @joshuay03, and apologies this sat for so long — it's exactly the right change, and the linked Rails PRs made the reasoning easy to follow. Rebased as #1692, with you as co-author.

Two things I added:

**The spec suite is converted too.** You noted you'd only covered non-test code. Since the whole point is that Ransack shouldn't touch the deprecated API, leaving the suite warning under `:deprecated` would have let regressions back in unnoticed. `condition_spec.rb` and `predicate_spec.rb` now use `adapter_class::ADAPTER_NAME` and `lease_connection`, and `lib/` has no `.connection` reference left at all.

**`bug_report_templates/` conflicted** because `main` deleted that directory in #1642. Kept deleted.

> since this gem currently supports 7.1 I've only relied on the connection pool method

That constraint is gone — the gemspec requires `>= 7.2` now, so `ActiveRecord::ConnectionHandling#with_connection` is available. Nothing in the current diff needs it, so I didn't add usage on that basis, but it's open if you want to take it further.

I verified on Active Record 7.2.2 (the oldest the gemspec allows) as well as 8.1.3 — `adapter_class`, `lease_connection`, `connection_pool.schema_cache` and `with_connection` all exist there, so no version guards are needed. 529 examples, 0 failures on both.

Closing in favour of #1692. This is also a concrete step toward @gregmolnar's point in #1640 about moving off Active Record's private APIs.

PR #1582, opened 2025-07-26
Fix PostgreSQL compatibility issue with DISTINCT and ORDER BY

### LINKED ISSUE #429: bug when sorting by association when using postresql with distinct(:true)
I know this is a known bug:
1. http://stackoverflow.com/questions/19748959/ransack-postgres-sort-on-column-from-associated-table-with-distinct-true
2. http://stackoverflow.com/questions/19910574/distinct-for-sorting-with-ransack

but is there a fix?

This code works:

```
def index
  @search = LogFile.search(params[:q])
  @log_files = @search.result(distinct: false).includes(:search_statistic).page(params[:page]).per(5)
end
```

This code does not:

```
def index
  @search = LogFile.search(params[:q])
  @log_files = @search.result(distinct: true).includes(:search_statistic).page(params[:page]).per(5)
end
```

Error:

```
PG::InvalidColumnReference: ERROR:  for SELECT DISTINCT, ORDER BY expressions must appear in select list
```

EduardMcfly:
## Problem

PostgreSQL requires that when using DISTINCT, all columns in the ORDER BY clause must also be present in the SELECT clause. This causes issues when using Ransack with DISTINCT queries that include sorting.

## Solution

This PR adds a new `DistinctSortsProcessor` class that:

- Automatically detects when a query uses DISTINCT with ORDER BY
- Adds necessary SELECT clauses for sort columns with unique aliases
- Updates the ORDER BY clause to use the aliased column names
- Maintains backward compatibility with existing queries

## Changes

- **New file**: `lib/ransack/distinct_sorts_processor.rb` - Core processor logic
- **Modified**: `lib/ransack/search.rb` - Integrates the processor into the result method
- **Modified**: `lib/ransack.rb` - Adds require for the new processor
- **New file**: `spec/ransack/distinct_sorts_processor_spec.rb` - Comprehensive test coverage

## Testing

The implementation includes extensive test coverage for:
- Different types of sort expressions (Arel nodes, strings)
- Complex sort strings with NULLS clauses
- Integration with existing Ransack functionality
- Edge cases and error conditions

## Compatibility

This change is fully backward compatible and only activates when both DISTINCT and ORDER BY are present in a query.

Closes #429


scarroll32:
> [!NOTE]

Thanks @EduardMcfly — closing this, and I want to be specific about why, because the problem it targets is real (PostgreSQL rejects `SELECT DISTINCT` when an `ORDER BY` expression isn't in the select list).

This PR adds ~1,300 lines — a new `DistinctSortsProcessor` and a general `SqlExpressionParser` that tokenises SQL strings — to handle that one case, and against current `main` it fails 10 specs. That's a lot of new surface, including a hand-written SQL parser, which is exactly the kind of fragile-by-construction code the 5.0 effort (#1640) is trying to remove rather than add. Rebasing it isn't realistic; it needs a smaller design.

A mergeable version would likely: work on Arel nodes rather than parsing SQL text (the sorts are already Arel `Ordering` nodes at the point `evaluate` sees them), add the missing order expressions to the projection only when `distinct: true` is set and the adapter is PostgreSQL, and come with specs on the PostgreSQL leg of CI. The existing note in `docs/docs/going-further/other-notes.md` ("Problem with DISTINCT selects") is where the documented workaround lives today.

If you'd like to take a run at that shape, please do open a fresh PR.

PR #1655, opened 2025-11-04
Add length predicates for string length search

### LINKED ISSUE #1492: Predicate to search by length of string
I have a rails API using ransack to filter results. I want to be able to query by the length of the string eg. I want to find all records where attribute a's length is less than a desired value.

name_length_lteq: 4 => should return all records with the name of 4 or less characters.

tatematsu-k:
## Summary
Implements predicates to search by string length as requested in #1492.

## Changes
- Added `length_eq`, `length_lt`, `length_lteq`, `length_gt`, `length_gteq` predicates
- These predicates wrap the attribute in LENGTH() or CHAR_LENGTH() function based on the database adapter
- Added comprehensive tests for all length predicates

## Usage
```ruby
# Find records where name length is less than or equal to 4
Person.ransack(name_length_lteq: 4).result

# Find records where name length equals 10
Person.ransack(name_length_eq: 10).result
```

## Database Support
- PostgreSQL/PostGIS: Uses `CHAR_LENGTH()`
- MySQL: Uses `CHAR_LENGTH()`
- SQLite and others: Uses `LENGTH()`

Closes #1492

### DIFF
--- a/lib/ransack/constants.rb
+++ b/lib/ransack/constants.rb
@@ -155,6 +155,31 @@ module Ransack
         type: :boolean,
         validator: proc { |v| BOOLEAN_VALUES.include?(v) },
         formatter: proc { |v| nil } }
+      ],
+      ['length_eq'.freeze, {
+        arel_predicate: 'eq'.freeze,
+        type: :integer
+        }
+      ],
+      ['length_lt'.freeze, {
+        arel_predicate: 'lt'.freeze,
+        type: :integer
+        }
+      ],
+      ['length_lteq'.freeze, {
+        arel_predicate: 'lteq'.freeze,
+        type: :integer
+        }
+      ],
+      ['length_gt'.freeze, {
+        arel_predicate: 'gt'.freeze,
+        type: :integer
+        }
+      ],
+      ['length_gteq'.freeze, {
+        arel_predicate: 'gteq'.freeze,
+        type: :integer
+        }
       ]
     ].freeze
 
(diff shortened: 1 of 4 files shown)


scarroll32:
> [!NOTE]

Thanks @tatematsu-k — clean implementation of a long-standing request (#1492 has been open since 2024). Rebased as #1697, with you as co-author.

Two changes:

**The adapter lookup used `ActiveRecord::Base.connection.adapter_name`.** That's the soft-deprecated call that #1692 is removing from `lib/` entirely, so merging this as written would have put one straight back. Switched to `adapter_class::ADAPTER_NAME`.

**Added Trilogy to the `CHAR_LENGTH` branch.** Trilogy is MySQL, so it has `CHAR_LENGTH`, but `when "Mysql2"` wouldn't match it and those users would silently get `LENGTH` instead. Since two of the three branches returned the same string, the `case` became a list.

Conflicts came from two things merged after you opened this: #1678's `present`/`blank` guard in `format_predicate` (both kept — guard first, then the length dispatch) and the `temperament` enum column in the spec schema (both column sets kept).

Verified all five predicates end-to-end, not just the SQL:

```
name_length_lteq: 3  =>  ["Jo"]
name_length_gt: 3    =>  ["Jonathan", ...]
```

Also added docs — five rows in the Search Matchers table and a short section on the `CHAR_LENGTH`/`LENGTH` difference.

Closing in favour of #1697.

Issue #940, opened 2018-08-07
does not work with null values using *_in matchers


siddushar:
i have requirement to search like below. where i can select multiple drop-down (ex: status as "active", "default") default status indicates that status column is empty or null.

select * from users where status in ("active","")

how can i achieve this using ransack..?

fernandoluizao:
I think you need to use *_in and *_blank, and then [merge](https://github.com/activerecord-hackery/ransack/wiki/Merging-Searches) the two together.

Or you can use [grouping](https://github.com/activerecord-hackery/ransack#grouping-queries-by-or-instead-of-and)

It would be nice if we have *_in_or_blank :smile:

agrberg:
We had this issue as well. We want to search for items assigned to Person A, B, or unassigned.

A bit of searching found [this page](https://qiita.com/t_fujise/items/2140ee4fe27859369fad) which has some examples of how to add this to Ransack as a new predicate. Copying the code here as well as you never know when links go offline 🙀 

```ruby
module Arel
  module Predications
    def eq_or_null(other)
      left  = eq(other)
      right = eq(nil)
      left.or(right)
    end

    def not_eq_or_null(other)
      left  = not_eq(other)
      right = eq(nil)
      left.or(right)
    end

    def gteq_or_null(other)
      left  = gteq(other)
      right = eq(nil)
      left.or(right)
    end

    def lt_or_null(other)
      left  = lt(other)
      right = eq(nil)
      left.or(right)
    end

    def in_or_null(other)
      left  = self.in(other)
      right = eq(nil)
      left.or(right)
    end
  end
end

Ransack.configure do |config|
  config.add_predicate 'eq_or_null', arel_predicate: 'eq_or_null'
  config.add_predicate 'not_eq_or_null', arel_predicate: 'not_eq_or_null'
  config.add_predicate 'gteq_or_null', arel_predicate: 'gteq_or_null'
  config.add_predicate 'lt_or_null', arel_predicate: 'lt_or_null'
  config.add_predicate 'in_or_null', arel_predicate: 'in_or_null', wants_array: true
end
```

scarroll32:
Would be a useful addition to the documentation.

Tobias-Knudsen:
I’ve noticed some inconsistent behavior when using the _in predicate in Ransack compared to ActiveRecord.

For example, in plain Rails:

```
Vehicle.where(brand: [nil])
[...]

Issue #524, opened 2015-03-20
searching with _in predicate if the argument is empty returns All records


tyronewilson:
I don't know if this issue has been brought up before but for me this is unexpected behaviour.

If I use a query something like `MyModel.search(parent_id_or_id_in: some_association.pluck(:my_model_id)).result` This works fine if I have some_association records. If I don't have any it will return ALL of the MyModel records instead of none. 

This is unexpected behaviour in my view and is a serious authorization issue for us as we are basing some user actions and permissions based on the results of these kinds of searches.

 If I search for `id_in: []` I get ALL records back, I would expect fan empty ActiveRecord::Relation under such circumstances. The same thing happens if I use `nil`

The following examples outline the problem:

1)  `SomeModel.search(id_in: []).result #all records returned` 
2)  `SomeModel.search(id_in: nil).result #all records returned`
3)  `SomeModel.search(id_in: SomeModel.none).result #all records returned`
4)  `SomeModel.search(id_in: {}).result #returns all records`

Some cases which are handled properly

1)  `SomeModel.search(id_in: 'bla').result #returns empty relation`
2)  `SomeModel.search(id_in: 3).result #returns empty relation`

As mentioned before, this has become a bit of a security issue for us and I hope you will see it the same way. I tried to see how one might fix it but not being familiar with such a big project as Ransack I wasn't sure.

Please let me know if we can get this fixed.

jonatack:
Hi @tyronewilson, you could populate the user search field input with a default value like 0 when you'd like to be certain the result with be filtered with a search.

P.S. - I'd recommend using the `#ransack` class method instead of `#search` for the reasons mentioned in the README.

tyronewilson:
Perhaps I am a bit slow. How does one specify the default value? I saw a trick in another issue where someone adds [-1] to all arrays for querying but I couldn't come up with a reliable way to do that without having a maintenance nightmare.

Perhaps you can provide me with an example of your meaning above @jonatack?

jonatack:
For reference, according this PR in Rails master https://github.com/rails/rails/pull/16924, this seems to be expected behavior (if I'm reading it accurately).

jonatack:
@tyronewilson you might have better luck on Stack Overflow where there is a larger community. I need to prioritize fixing issues with the time I have for Ransack. Thanks for your understanding!

tyronewilson:
[...]

Issue #1640, opened 2025-09-25
Release 6.0.0


scarroll32:
_Please let me know if you would like to help with this significant effort, you can ping me or else self-assign any unassigned issues with Milestone 6.0.0._

Ransack is ~~almost at 100 million downloads~~ at [108 million downloads](https://rubygems.org/gems/ransack), but it has a lot of identified bugs and the sprawling codebase is challenging. The current release of Ransack is 5.0.0 (2026-09-21) and the next major version will be 6.0.0. This will be an effort to clean up the many bugs in Ransack's backlog and refactor the codebase so it is easier to extend or fix bugs in future and it will be a release with potential breaking changes. 

Candidate `6.0.0` issues are marked with the [6.0.0 milestone](https://github.com/activerecord-hackery/ransack/issues?q=is%3Aissue%20state%3Aopen%20milestone%3A6.0.0) and linked to this issue once confirmed.

To test against this release as it unfolds, please point your Gemfile to `main` (the `v5.0.0` branch was consolidated into `main` in #1647):

```
gem 'ransack', :github => 'activerecord-hackery/ransack', :branch => 'main'
```

New features such as full text search can be added after 6.0.0 depending on community interest and involvement; the candidate issues are currently in the [Backlog milestone](https://github.com/activerecord-hackery/ransack/issues?q=is%3Aissue%20state%3Aopen%20milestone%3ABacklog).

### High level tasks 
- [x] Improve test coverage
- [ ] Simplify documentation
- [ ] Explicitly define Ransack's boundries in code and documentation (adapters and gem interactions)
- [ ] Refactor the codebase 
- [ ] Cleanup outstanding bugs

### Issues in this release

Every open issue was re-tested against 5.0.0 on 2026-09-21. Issues that turned out to be already fixed, not bugs, or feature requests were closed or moved to the Backlog milestone with an explanation on each; the ones below reproduce and are being fixed.

**Compatibility and structure**
- [ ] #1686 Drop Ruby 3.1
- [ ] #1625 Remove the `Polyamorous` and `Adapters` namespaces
- [ ] #1631 Refactor the module and file structure
- [ ] #1639 Explicitly support connection adapters
- [ ] #1633 Remove the PostGIS dependency
- [ ] #1680 Replace Docusaurus with Just the Docs
- [ ] #1613 Simplify documentation

**Bugs: small, self-contained**
- [ ] #1009 `predicate_name:` long key is ignored
[...]

PR #1657, opened 2025-11-11
Allow nil values in array

Tobias-Knudsen:
I’ve noticed some inconsistent behavior when using the _in predicate in Ransack compared to ActiveRecord.

For example, in plain Rails:

```
Vehicle.where(brand: [nil])
# => SELECT "vehicles".* FROM "vehicles" WHERE "vehicles"."brand" IS NULL
```

```
Vehicle.where(brand: [nil, 'AUDI'])
# => SELECT "vehicles".* FROM "vehicles" WHERE ("vehicles"."brand" = $1 OR "vehicles"."brand" IS NULL)
```

We use Ransack in our API. Since Rails automatically strips any nil values from an array in params, it's not possible for us to use the standard `_in` predicate in our API. To accomplish this I've introduced a custom predicate `_in_or_null` as described here: https://github.com/activerecord-hackery/ransack/issues/940#issuecomment-1410704480

This allows me to perform the second example by using `brand_in_or_null: [ "AUDI" ]`, however if I would use `brand_in_or_null: []`, I would expect this to only give me the vehicles where the brand is null, however the current logic deletes the param and returns all records. 

I've changed the logic in the Search class initializer, where it now allows an array with nil values.

Possibly related to: https://github.com/activerecord-hackery/ransack/issues/524

### DIFF
--- a/lib/ransack/search.rb
+++ b/lib/ransack/search.rb
@@ -26,7 +26,7 @@ module Ransack
       if params.is_a? Hash
         params = params.dup
         params = params.transform_values { |v| v.is_a?(String) && strip_whitespace ? v.strip : v }
-        params.delete_if { |k, v| [*v].all?{ |i| i.blank? && i != false } }
+        params.delete_if { |k, v| [*v].all?{ |i| i.blank? && i != false && !i.nil? } }
       else
         params = {}
       end
--- a/spec/ransack/search_spec.rb
+++ b/spec/ransack/search_spec.rb
@@ -75,6 +75,12 @@ module Ransack
         Search.new(Person, name_eq_any: ['foobar'])
       end
 
+      it 'keeps conditions with a nil value in an array before building' do
+        expect_any_instance_of(Search).to receive(:build)
+        .with({ 'name_in' => [nil] })
+        Search.new(Person, name_in: [nil])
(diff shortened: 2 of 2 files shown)


scarroll32:
> [!NOTE]

Thanks @Tobias-Knudsen. The asymmetry with Active Record is real — `where(brand: [nil])` produces `IS NULL`, while Ransack dropped the condition entirely — and the custom-predicate workaround from #940 is unusable without this.

The guard is correctly scoped: `[*nil]` is `[]` and `[].all?` is still `true`, so a bare `name_eq: nil` is dropped exactly as before. Only an explicit `nil` **inside an array** now survives, which is the case #940 and #524 are about.

Verified locally merged into current `main` — Rails 8.1.3 / Ruby 3.4.9 / SQLite: **527 examples, 0 failures, 1 pending**.

Merging with an admin override: the branch predates #1667, so the Rails 8.1 / Ruby 3.3–4.0 required contexts can't report on it. All 15 checks it defines are green.

Worth noting for the 5.0.0 release notes (#1640): `name_in: [nil, '']` is now kept where it was previously dropped. Low blast radius in practice, since Rails strips `nil` from array params by default, but it is a behaviour change.

PR #1660, opened 2025-12-13
Fix wildcard escaping for SQLite and other databases

### LINKED ISSUE #1581: LIKE predicates do not escape values in SQLite and other non-MySQL/PostgreSQL RDBMS
**Summary**

Ransack's `*_cont`, `*_not_cont`, and similar LIKE-based predicates correctly escape special characters in MySQL, PostgreSQL, and PostGIS. However, in SQLite and other RDBMS, escaping is skipped, which can lead to unintended matches when values contain `%` or `_`.

**Steps to Reproduce**

Run a Ransack query using `*_cont: '%_'` against a dataset that includes values with and without underscores. In SQLite, the query matches records that do not contain underscores, due to lack of escaping.

**Expected Behavior**

The query should escape `%` and `_` so that they are treated as literal characters in all supported databases.

**Actual Behavior**

In SQLite, the escaping is skipped, resulting in unintended matches.

**Environment**

- Ruby: 3.4.5  
- Rails: 8.0.2  
- Ransack: commit 90a602a992f4e3a438d63e0fc2f7e1a0fb3acab0  
- sqlite3-ruby: 2.7.3

**Code Reference**

`escape_wildcards` skips escaping for non-MySQL/PostgreSQL adapters. https://github.com/activerecord-hackery/ransack/blob/90a602a992f4e3a438d63e0fc2f7e1a0fb3acab0/lib/ransack/constants.rb#L162-L174

**Test Reference**

`predicate_spec.rb` asserts that escaping is not applied in SQLite. https://github.com/activerecord-hackery/ransack/blob/90a602a992f4e3a438d63e0fc2f7e1a0fb3acab0/spec/ransack/predicate_spec.rb#L158-L176

**Suggested Fix**

Use `ActiveRecord::Base.sanitize_sql_like` to escape LIKE values consistently across all adapters.
https://api.rubyonrails.org/v8.0.2/classes/ActiveRecord/Sanitization/ClassMethods.html#method-i-sanitize_sql_like

**Reproduction Code**

```ruby
# frozen_string_literal: true

# Create Gemfile dynamically
File.write('Gemfile', <<~GEMFILE)
  source 'https://rubygems.org'
  gem 'rails', '8.0.2'
  gem 'ransack', github: 'activerecord-hackery/ransack', ref: '90a602a992f4e3a438d63e0fc2f7e1a0fb3acab0'
  gem 'sqlite3', '2.7.3'
GEMFILE

# Install dependencies
system 'bundle install'

# Setup Bundler
require 'bundler'
Bundler.setup(:default)

# Require necessary libraries
require 'active_record'
require 'logger'
require 'ransack'

# Setup in-memory SQLite DB
ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: ':memory:')
ActiveRecord::Base.logger = Logger.new($stdout)

# Define schema
  create_table :products, force: true do |t|
    t.string :name
  end
end

# Define model
class Product < ActiveRecord::Base
  def self.ransackable_attributes(_auth_object = nil)
    %w[name]
  end
end

# Seed data
Product.create!(name: '100% Pure')
Product.create!(name: '50_50 Blend')
Product.create!(name: 'Back\\Slash')
Product.create!(name: 'NormalText')

# Ransack query with special characters
q = Product.ransack(name_cont: '%_')
puts "\nGenerated SQL:"
puts q.result.to_sql
puts "\nResults:"
puts q.result.map(&:name)
```

**Output (Excerpt)**

```
Generated SQL:
SELECT "products".* FROM "products" WHERE "products"."name" LIKE '%%_%'

Results:
100% Pure
50_50 Blend
Back\Slash
NormalText
```

heyjosephme:
Fixes #1581

## Problem
SQLite and non-MySQL/PostgreSQL databases weren't escaping SQL wildcards (%, _, \) in LIKE queries, causing unintended matches.

## Solution
Replace database-specific escaping with Rails' `sanitize_sql_like` method which works for all databases.

## Changes
- Simplified `escape_wildcards` method in `lib/ransack/constants.rb`
- Updated test expectations for SQLite in `spec/ransack/predicate_spec.rb`
- Changed from 13 lines of regex to 1 line using Rails built-in

## Testing
- ✅ All tests pass (8 examples, 0 failures)
- ✅ Wildcard escaping now works correctly for SQLite

---

This is my first contribution to Ransack. Looking forward to feedback!

### DIFF
--- a/lib/ransack/constants.rb
+++ b/lib/ransack/constants.rb
@@ -159,18 +159,10 @@ module Ransack
     ].freeze
 
   module_function
-    # replace % \ to \% \\
+    # replace % \ _ to \% \\ \_
+    # Uses Rails' sanitize_sql_like which works for all databases
     def escape_wildcards(unescaped)
-      case ActiveRecord::Base.connection.adapter_name
-      when "Mysql2".freeze
-        # Necessary for MySQL
-        unescaped.to_s.gsub(/([\\%_])/, '\\\\\\1')
-      when "PostGIS".freeze, "PostgreSQL".freeze
-        # Necessary for PostgreSQL
-        unescaped.to_s.gsub(/([\\%_.])/, '\\\\\\1')
-      else
-        unescaped
-      end
+      ActiveRecord::Base.sanitize_sql_like(unescaped.to_s)
     end
   end
 end
--- a/spec/ransack/predicate_spec.rb
+++ b/spec/ransack/predicate_spec.rb
@@ -163,7 +163,7 @@ module Ransack
         when "Mysql2"
           /`people`.`name` LIKE '%\\\\%.\\\\_\\\\\\\\%'/
         else
-         /"people"."name" LIKE '%%._\\%'/
+         /"people"."name" LIKE '%\\%.\\_\\\\%'/
         end) do
         subject { @s }
       end
@@ -183,7 +183,7 @@ module Ransack
         when  "Mysql2"
           /`people`.`name` NOT LIKE '%\\\\%.\\\\_\\\\\\\\%'/
         else
-         /"people"."name" NOT LIKE '%%._\\%'/
+         /"people"."name" NOT LIKE '%\\%.\\_\\\\%'/
         end) do
         subject { @s }
       end
@@ -205,7 +205,7 @@ module Ransack
         when "Mysql2"
           /LOWER\(`people`.`name`\) LIKE '%\\\\%.\\\\_\\\\\\\\%'/
         else
-         /LOWER\("people"."name"\) LIKE '%%._\\%'/
+         /LOWER\("people"."name"\) LIKE '%\\%.\\_\\\\%'/
         end) do
         subject { @s }
       end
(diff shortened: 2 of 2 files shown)


scarroll32:
> [!NOTE]

Thank you @heyjosephme — and welcome, this is a good first contribution. You identified the right bug and the right place to fix it.

I've picked this up and finished it in #1682, with you as co-author.

The reason this couldn't merge as-is is worth explaining, because it's the interesting part of #1581. Switching to `sanitize_sql_like` escapes on every adapter, which is the necessary first half — but it fails four existing specs, and those failures are real:

**MySQL and PostgreSQL treat `\` as the default `LIKE` escape character. SQLite has no default at all.** So on SQLite, escaping `a_c` to `a\_c` doesn't make the underscore literal — it produces a pattern looking for a literal backslash, which now matches neither `a_c` nor `abc`. The bug changes shape rather than going away. That's what those four red specs were telling us.

The complete fix needs the escape character declared in the SQL as well:

```sql
WHERE "people"."name" LIKE '%a\_c%' ESCAPE '\'
```

Arel's `matches` takes that as its second argument, so #1682 keeps your uniform escaping and threads the escape character through `Condition#format_predicate`. It also covers the `_any`/`_all` compound forms, which take the same argument and would otherwise have kept the old behaviour.

Two things your change made possible that are worth calling out: dropping the adapter switch removes a call to the deprecated `ActiveRecord::Base.connection` from the formatting path, and it let the four blocks of nested-escape regex in the specs collapse into something readable.

Closing in favour of #1682. Thanks again — please do send more.

Issue #722, opened 2016-09-16
predicate containing empty string results in ignored query clause


chrismagnacca:
When passing a hash of query parameters to ransack, the documentation states that nil predicate values will be ignored. However, empty string values for predicates are also ignored. For example,

``` ruby
{
  "nameIn": [""],
  "addressIn": ["123 Fake St."]
}
```

If we apply the query params above to a fictional User model, the generated query would be:

``` sql
SELECT `users`.* FROM `users`  WHERE ((`users`.`address` IN ('123 Fake St.'));
```

Additionally, the same behavior occurs when passing empty string into Eq clauses. Based on the documentation, I would expect the following query to be generated:

``` sql
SELECT `users`.* FROM `users`  WHERE ((`users`.`address` IN ('123 Fake St.') AND `users`.`name` IN (''));
```

If this is not the desired case, then including this in the documentation would be preferred. If a user was specifying an Eq clause against Ids and (through some method) empty string ends up as the predicate value, the result set would go from 1 (in the case of Eq) to N (all records, since the idEq clause would be dropped for empty string).

Tobias-Knudsen:
Bump

PR #1664, opened 2026-02-01
Allow empty string and empty array as valid search values

pekopekopekopayo:
Ransack currently discards empty strings and empty arrays from search params silently.

This might be intentional, but users may not be aware of this behavior or may not want it. If a user passes an empty array, they probably expect no results to be returned. If they pass an empty string, they probably want to find records where the field is actually empty.

Currently, passing empty values causes a full table scan because all conditions are removed. This could potentially lead to OOM issues (unless pagination is in place).

With this change:
  - `name_eq: ''` generates `WHERE name = ''`
  - `id_in: []` generates `WHERE 1=0` (returns no results)
  - `name_eq: nil` is still ignored (existing behavior preserved)

Related: #1592, #722

### DIFF
--- a/lib/ransack/predicate.rb
+++ b/lib/ransack/predicate.rb
@@ -37,8 +37,7 @@ module Ransack
       @arel_predicate = opts[:arel_predicate]
       @type = opts[:type]
       @formatter = opts[:formatter]
-      @validator = opts[:validator] ||
-        lambda { |v| v.respond_to?(:empty?) ? !v.empty? : !v.nil? }
+      @validator = opts[:validator] || lambda { |v| !v.nil? }
       @compound = opts[:compound]
       @wants_array = opts.fetch(:wants_array,
         @compound || Constants::IN_NOT_IN.include?(@arel_predicate))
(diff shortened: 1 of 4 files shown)


scarroll32:
> [!NOTE]

Thanks @pekopekopekopayo — this is a well-argued PR and the point about the full table scan is a good one. I've reworked it in #1683, with you as co-author.

To answer your opening question directly: yes, the current behaviour is intentional, but not for the reason it might look like. A blank text input in an HTML form posts `""`. Ransack dropping it is what makes a search form with empty fields return every record instead of none. Flipped unconditionally, every untouched field in every Ransack form becomes `WHERE column = ''` and those forms silently return nothing.

That doesn't make your case wrong — it makes it a different caller. For a JSON API an empty value really is an explicit filter, and `id_in: []` matching everything is indefensible there. Both behaviours are correct, so #1683 makes it `config.ignore_blank_values`, defaulting to today's behaviour.

Your three-part diagnosis carried over almost unchanged, and it was the useful part — the option only works because all three places agree:

- the params filter in `Search#initialize`
- the default predicate validator
- `Predicate#validate`, for the empty-array-with-`wants_array` case you spotted

One thing I changed: the validator now reads the option at search time rather than capturing it when the predicate is constructed. Predicates are registered at load, before `config/initializers/ransack.rb` runs, so a captured value would have made the setting silently do nothing.

Closing in favour of #1683. Thanks for pushing on this — #722 has been open since 2016.

Issue #1644, opened 2025-09-26
Non-boolean values passed to boolean predicates are no longer cast


dennmart:
The changes made in #1559 changed how values are casted for some/all boolean predicates like `not_null`.

Before the changes from the PR, when passing a non-boolean value to a boolean predicate (for example: `name_not_null = 0`), the `Ransack::Nodes::Condition#casted_values_for_attribute` method would call `Ransack::Nodes::Value#cast` using the predicate or attribute type. When the predicate type is boolean, the value would be cast to a proper boolean.

The changes in the PR just pass the value directly without any casting, so when calling the `arel_predicate` proc for the predicate, it will evaluate to `true` because `0` is truthy. Using a boolean value (like `name_not_null = false`) works as expected.

I'm not sure if this is a bug and these values should be properly cast, or if we're expected with Ransack 4.4.0 to use boolean values for these types of predicates.

Original issue https://github.com/activerecord-hackery/ransack/issues/1555

pioz:
Same issue here. Version 4.4.0 breaks all my code. Is this a bug or an intentional breaking change?

scarroll32:
@matheussilvasantos what are your thoughts on this issue?

Thank you for the report @dennmart and @pioz - this does appear to be a bug. Please also see the request in [this issue](https://github.com/activerecord-hackery/ransack/issues/1640#issue-3452854317) for a large number of changes coming to Ransack.

scarroll32:
> Same issue here. Version 4.4.0 breaks all my code. Is this a bug or an intentional breaking change?

Not intentional.

matheussilvasantos:
@deivid-rodriguez can help more than me in this case.

kzkn:
Same issue for timestamp column.
```
app(dev)> Time.zone
=> #<ActiveSupport::TimeZone:0x00007fb8e1a708b8 @name="Tokyo", @tzinfo=#<TZInfo::DataTimezone: Asia/Tokyo>, @utc_offset=nil>

# 4.3.0
app(dev)> puts User.ransack(created_at_gteq: '2025-09-29T00:00').result.to_sql
SELECT "users".* FROM "users" WHERE "users"."created_at" >= '2025-09-28 15:00:00'

# 4.4.0
app(dev)> puts User.ransack(created_at_gteq: '2025-09-29T00:00').result.to_sql
SELECT "users".* FROM "users" WHERE "users"."created_at" >= '2025-09-29 00:00:00'
```

[...]

PR #1665, opened 2026-02-04
support enum

### LINKED ISSUE #1555: Why Doesn’t Ransack Support Rails Enums Properly?

pekopekopekopayo:
Fixes #1555
Related: #1644

## Problem

When searching enum attributes by key (e.g., `temperament_eq: 'choleric'`), Ransack identified the column type as `:integer` and cast the value via `String#to_i`, resulting in `'choleric'.to_i == 0`. This produced incorrect queries like `WHERE temperament = 0` instead of `WHERE temperament = 2`.

## Solution

In Rails, `Person.where(temperament: 'choleric')` and `Person.where(temperament: 2)` both produce the same query — ActiveRecord handles the mapping internally. Ransack should behave the same way. To achieve this, `Attribute#type` now returns `:enum` for enum attributes, and `Value#cast(:enum)` passes the raw value through to ActiveRecord instead of attempting its own conversion.

A regression test for [#1644](https://github.com/activerecord-hackery/ransack/issues/1644) has also been included to ensure boolean predicate casting remains unaffected.

## Changes

- `Attribute#type` — returns `:enum` when the attribute is a defined enum
- `Value#cast(:enum)` — returns raw value, delegating conversion to ActiveRecord
- Added enum search specs (key and integer value)
- Added regression test for #1644 (boolean predicate casting)

### DIFF
--- a/lib/ransack/nodes/attribute.rb
+++ b/lib/ransack/nodes/attribute.rb
@@ -31,6 +31,8 @@ module Ransack
       def type
         if ransacker
           ransacker.type
+        elsif enum?
+          :enum
         else
           context.type_for(self)
         end
@@ -52,6 +54,14 @@ module Ransack
 
       def inspect
         "Attribute <#{name}>"
+      end
+
+      private
+
+      def enum?
+        bound? &&
+        klass.respond_to?(:defined_enums) &&
+        klass.defined_enums.key?(attr_name.to_s)
       end
 
     end
--- a/lib/ransack/nodes/value.rb
+++ b/lib/ransack/nodes/value.rb
@@ -24,6 +24,8 @@ module Ransack
 
       def cast(type)
         case type
+        when :enum
+          value
         when :date
           cast_to_date(value)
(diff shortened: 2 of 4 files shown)


pekopekopekopayo:
@deivid-rodriguez
@scarroll32

Could you please review this?

PR #1647, opened 2025-09-29
Consolidate the v5.0.0 branch into main

scarroll32:
Part of #1640 (does not close it).

`v5.0.0` and `main` had diverged, and PRs were arriving against both branches. This merges the two so `main` becomes the single line of development for the 5.0.0 effort.

On `v5.0.0` but not `main`:
- Add pending spec capturing issue #1553 (#1672)
- Fix scopes inside groupings being silently dropped (#1339, #1490) (#1673)
- Add DeepWiki badge (#1643)
- Update gem release docs (#1646)

On `main` but not `v5.0.0`:
- support enum (#1665)
- Add CI with Rails 8.1 and Ruby 3.3.x, 3.4.x and 4.0.1 (#1667)
- Revise CHANGELOG for versioning and support updates (#1648)

The merge is conflict-free. Verified locally against Rails 8.1 / Ruby 3.4 on SQLite: **517 examples, 0 failures, 1 pending** (the intentionally-pending #1553 spec added in #1672).

After this lands, `v5.0.0` is fast-forwarded to match `main`, and open PRs targeting `v5.0.0` are retargeted to `main`.

PR #1675, opened 2026-06-01
Fix low-level c: API silently dropping conditions with Array envelope hashes (#1150)

### LINKED ISSUE #1150: Grouping doesn't work when "a" is given as an array
```ruby
[18] pry(main)> Contact.ransack(g: [{c: [{a: [{name: "name"}], v: [{value: "Kasper"}], p: "cont"}], m: "and"}]).result.to_sql
=> "SELECT \"contacts\".* FROM \"contacts\""
```

```ruby
[17] pry(main)> Contact.ransack(g: [{c: [{a: {"0"=>{name: "name"}}, v: [{value: "Kasper"}], p: "cont"}], m: "and"}]).result.to_sql
=> "SELECT \"contacts\".* FROM \"contacts\" WHERE \"contacts\".\"name\" ILIKE '%{\"value\"=>\"Kasper\"}%'"
```

ryoya1122:
Closes #1150.

## Problem

The low-level conditions API (`c: [{a:, v:, p:}]`) accepts `a:` (attributes) and `v:` (values) in two equivalent forms:

- Hash form: `a: {'0' => {name: 'foo'}}`, `v: {'0' => {value: 'bar'}}`
- Array form: `a: ['foo']`, `v: ['bar']`

However when the Array form contained the same envelope hash that the Hash form unwraps, the parser did the wrong thing.

**Bug 1 — `a:` Array of `{name: ...}` envelopes silently drops the condition:**

```ruby
Contact.ransack(g: [{ c: [{ a: [{ name: 'name' }], v: [{ value: 'Kasper' }], p: 'cont' }], m: 'and' }]).result.to_sql
# => SELECT "contacts".* FROM "contacts"
#    ^ no WHERE clause
```

**Bug 2 — `v:` Array of `{value: ...}` envelopes leaks the whole hash into the SQL:**

```ruby
Contact.ransack(g: [{ c: [{ a: { '0' => { name: 'name' } }, v: [{ value: 'Kasper' }], p: 'cont' }], m: 'and' }]).result.to_sql
# => SELECT "contacts".* FROM "contacts" WHERE "contacts"."name" LIKE '%{"value" => "Kasper"}%'
```

The Hash form of both `a:` and `v:` worked correctly, so the asymmetry was the bug.

## Cause

`Condition#attributes=` and `#values=` switch on `Array`/`Hash`. The Hash branches unwrap each entry's `:name` / `:value`, but the Array branches assumed every element was a plain name or plain value. When an envelope hash was supplied, it was passed straight through:

- `attributes=` Array branch: `Attribute.new(@context, {name: 'name'}, [])` was built, failed `valid?`, and was dropped.
- `values=` Array branch: `Value.new(@context, {value: 'Kasper'})` was built, which later renders the hash via `to_s` inside the SQL.

## Fix

In the Array branch of each setter, detect envelope hashes (with `:name` or `:value` keys, either symbol or string) and extract the same fields the Hash branch already does. Plain Array elements (the canonical `a: ['name']`, `v: ['Ernie']` form used throughout existing specs) continue to work unchanged.

## Compatibility

The Array branches keep the existing behavior for every shape **except** the two envelope forms documented as broken in #1150:

| `a:` / `v:` element shape          | Before     | After     |
| ---------------------------------- | ---------- | --------- |
| `"foo"` (plain name / value)       | works      | works     |
| `{name: ...}` / `{value: ...}`     | **broken** | **fixed** |
| `{"name" => ...}` (string key)     | broken     | fixed     |
| `{name:, ransacker_args:}`         | broken     | fixed     |
| `{other_key: ...}` (no envelope)   | unchanged  | unchanged |
| `nil`                              | unchanged  | unchanged |

The envelope-detection check is intentionally narrow (`Hash` *and* has a `:name`/`:value` key) so that arbitrary Hash values — e.g. searches against a JSON column where the value itself is a Hash — keep flowing through Arel unchanged.

CI is green on the full matrix (SQLite / PostgreSQL / PostGIS / MySQL × Rails 7.2-stable / 8.0-stable × Ruby 3.1.4 / 3.2.2).

## Tests

Three new specs under `Ransack::Search#build > low-level c: API with Array of envelope hashes` in `spec/ransack/search_spec.rb`:

- extracts `:name` from each attribute hash inside an Array (red before, green after)
- extracts `:value` from each value hash inside an Array (red before, green after)
- still accepts plain Array of names/values, i.e. canonical form (green before and after — guards against regression)

Full suite: `515 examples, 0 failures, 1 pending` on `v5.0.0` + this branch (SQLite, ActiveRecord 7.2.3.1, Ruby 3.4.9).

Targets `v5.0.0` per #1640.

### DIFF
--- a/lib/ransack/nodes/condition.rb
+++ b/lib/ransack/nodes/condition.rb
@@ -81,8 +81,13 @@ module Ransack
       def attributes=(args)
         case args
         when Array
-          args.each do |name|
-            build_attribute(name)
+          args.each do |attr|
+            if attr.is_a?(Hash) && (attr.key?(:name) || attr.key?("name"))
+              attr = attr.with_indifferent_access
+              build_attribute(attr[:name], attr[:ransacker_args])
+            else
+              build_attribute(attr)
+            end
           end
         when Hash
           args.each do |index, attrs|
@@ -104,8 +109,10 @@ module Ransack
         case args
         when Array
           args.each do |val|
-            val = Value.new(@context, val)
-            self.values << val
+            if val.is_a?(Hash) && (val.key?(:value) || val.key?("value"))
+              val = val.with_indifferent_access[:value]
+            end
+            self.values << Value.new(@context, val)
           end
         when Hash
           args.each do |index, attrs|
--- a/spec/ransack/search_spec.rb
+++ b/spec/ransack/search_spec.rb
@@ -242,6 +242,38 @@ module Ransack
         .to eq [Nodes::Condition, Nodes::Condition]
       end
 
+      # Regression test for https://github.com/activerecord-hackery/ransack/issues/1150
+      # The low-level `c:` API previously only worked when `a:` and `v:` were given
+      # in their Hash form (`{"0" => {name: ...}}`). When supplied as Arrays of
+      # attribute / value envelopes (`[{name: ...}]`, `[{value: ...}]`) the parser
+      # silently dropped the attribute or treated the entire envelope hash as the
+      # value.
+      context 'low-level c: API with Array of envelope hashes' do
+        it 'extracts :name from each attribute hash inside an Array' do
+          s = Search.new(Person,
+            c: [{ a: [{ name: 'name' }], p: 'cont', v: [{ value: 'Ernie' }] }]
+          )
+          field = "#{quote_table_name('people')}.#{quote_column_name('name')}"
+          expect(s.result.to_sql).to match(/#{field} I?LIKE '%Ernie%'/)
+        end
+
+        it 'extracts :value from each value hash inside an Array' do
+          s = Search.new(Person,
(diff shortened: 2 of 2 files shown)


scarroll32:
> [!NOTE]

Thanks @ryoya1122. The asymmetry between the Hash and Array forms of `a:`/`v:` is exactly the bug in #1150, and the compatibility table showing that arbitrary Hash values (e.g. JSON-column searches) still pass through untouched is what makes this safe to take.

Verified locally merged into current `main` — Rails 8.1.3 / Ruby 3.4.9 / SQLite: **520 examples, 0 failures, 1 pending**.

Merging with an admin override: the branch predates #1667 so the Rails 8.1 / Ruby 3.3–4.0 required contexts can't report on it, while every job it does define is green across all four adapters.

PR #1676, opened 2026-06-01
Make search_form_* helpers honour per-search search_key (#1118)

### LINKED ISSUE #1118: Use `search_key` from the search context for `search_form_for`
Hi there.

For `sort_link` and `sort_url` [`search_key` is taken from context](https://github.com/activerecord-hackery/ransack/blob/f663d64c507129440b3d2cbff64289faac426b2e/lib/ransack/helpers/form_helper.rb#L134) but for `search_form_for` [from the `Ransack` options](https://github.com/activerecord-hackery/ransack/blob/f663d64c507129440b3d2cbff64289faac426b2e/lib/ransack/helpers/form_helper.rb#L30) and this is not correct.

ryoya1122:
Closes #1118.

## Problem

When a `search_key:` is passed to `Model.ransack`, the form-building helpers ignore it. The inputs are emitted under the global default key (`q[...]`) regardless of the override:

```ruby
q = Person.ransack({}, search_key: :people_search)
search_form_for(q) { |f| f.text_field :name_eq }
# => <input type="text" name="q[name_eq]" id="q_name_eq" />
#                       ^^^ should be people_search[name_eq] / people_search_name_eq
```

The same applies to `search_form_with` and `turbo_search_form_for`. `sort_link` and `sort_url` already read the key from the search's context, so the form path is the outlier.

## Cause

The two private finalisers used the global default instead of asking the search:

```ruby
def finalize_form_options(options, html_options)
  options[:as] ||= Ransack.options[:search_key]   # <- global, ignores per-search override
  ...
end

def finalize_form_with_options(options, html_options)
  options[:scope] ||= Ransack.options[:search_key] # <- same
  ...
end
```

The reporter linked both lines in #1118.

## Fix

`finalize_form_options` / `finalize_form_with_options` now take the `search` and read `search.context.search_key`, the same source `sort_link` / `sort_url` already use (see `lib/ransack/helpers/form_helper.rb#L200` and `#L249`).

`Context#search_key` is initialised as `options[:search_key] || Ransack.options[:search_key]`, so the global default still applies when no per-search override is given.

## Compatibility

| `Model.ransack` call                                          | Before                              | After                                  |
| ------------------------------------------------------------- | ----------------------------------- | -------------------------------------- |
| `Person.ransack(...)`                                         | inputs under `q[...]` (global default) | inputs under `q[...]` (unchanged)      |
| `Person.ransack(...)` after `Ransack.configure { ... :example }` | inputs under `example[...]` (global) | inputs under `example[...]` (unchanged) |
| `Person.ransack(..., search_key: :people_search)`             | inputs under `q[...]` ❌ (#1118)     | inputs under `people_search[...]` ✓    |
| `options[:as]` / `options[:scope]` explicitly set by caller   | wins via `||=`                      | wins via `||=` (unchanged)             |

The `finalize_form_*` methods are `private`, so the signature change is internal.

## Tests

Two new specs in `spec/ransack/helpers/form_helper_spec.rb`, alongside the existing global-default tests, both red before the fix and green after:

- `#search_form_for with per-search search_key` -> expects `people_search_name_eq`
- `#search_form_with with per-search search_key` -> expects `people_search[name_eq]`

Full suite: `514 examples, 0 failures, 1 pending` on `v5.0.0` + this branch (SQLite, ActiveRecord 7.2.3.1, Ruby 3.4.9).

Targets `v5.0.0` per #1640.

### DIFF
--- a/lib/ransack/helpers/form_helper.rb
+++ b/lib/ransack/helpers/form_helper.rb
@@ -10,7 +10,7 @@ module Ransack
         search = extract_search_and_set_url(record, options, 'search_form_for')
         options[:html] ||= {}
         html_options = build_html_options(search, options, :get)
-        finalize_form_options(options, html_options)
+        finalize_form_options(search, options, html_options)
         form_for(record, options, &proc)
       end
 
@@ -30,7 +30,7 @@ module Ransack
         search = extract_search_and_set_url(record, options, 'search_form_with')
         options[:html] ||= {}
         html_options = build_html_options(search, options, :get)
-        finalize_form_with_options(options, html_options)
+        finalize_form_with_options(search, options, html_options)
         form_with(model: search, **options, &proc)
       end
 
@@ -48,7 +48,7 @@ module Ransack
         turbo_options = build_turbo_options(options)
         method = options.delete(:method) || :post
         html_options = build_html_options(search, options, method).merge(turbo_options)
-        finalize_form_options(options, html_options)
+        finalize_form_options(search, options, html_options)
         form_for(record, options, &proc)
       end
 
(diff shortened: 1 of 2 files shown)


scarroll32:
> [!NOTE]

Thanks @ryoya1122. Good catch that `sort_link`/`sort_url` already read `search.context.search_key` while the form helpers went straight to the global — that asymmetry was the whole of #1118. Reading from the context also means the global default still applies when no override is given, so existing apps are unaffected.

Verified locally merged into current `main` — Rails 8.1.3 / Ruby 3.4.9 / SQLite: **519 examples, 0 failures, 1 pending**.

Merging with an admin override: the branch predates #1667 so the Rails 8.1 / Ruby 3.3–4.0 required contexts can't report on it; every job it does define is green.

PR #1679, opened 2026-06-10
Fix deprecation warning and correctness bug when using arel_extensions >= 2.2

javieromar7:
## Problem

When `arel_extensions >= 2.2` is present, `extract_correlated_key` triggers a deprecation warning on every query involving correlated joins:

```
DEPRECATION WARNING: `ArelExtensions::Attributes#==` is now deprecated. Use `.eq` instead.
(called from extract_correlated_key at lib/ransack/adapters/active_record/context.rb:215)
```

Beyond the warning, there is a **silent correctness bug**: `ArelExtensions >= 2.2` overrides `==` on Arel attributes to return an `Arel::Nodes::Equality` SQL predicate node — not a boolean. This means the `if join_root.left == pk` check is always truthy (an Arel node is never `nil`/`false`), so `join_root.right` is returned unconditionally, regardless of whether `join_root.left` is actually the primary key.

## Root cause

```ruby
# arel_extensions/lib/arel_extensions/attributes.rb
def ==(other)
  deprecated 'Use `.eq` instead.'
  Arel::Nodes::Equality.new self, Arel.quoted(other, self)  # returns an Arel node, not a boolean
end
```

## Fix

Replace `==` with `eql?` for the primary key identity check. `Arel::Attributes::Attribute#eql?` performs value-based comparison (name + relation) without going through the ArelExtensions override.

```ruby
# Before
if join_root.left == pk
# After
if join_root.left.eql?(pk)
```

## Verification

```ruby
pk  = User.arel_table[:id]
pk2 = User.arel_table[:id]

pk.equal?(pk2)  # => false (different objects)
pk.eql?(pk2)    # => true  (same name + relation)
pk == pk2       # => Arel::Nodes::Equality (always truthy — the bug)
```

### DIFF
--- a/lib/ransack/adapters/active_record/context.rb
+++ b/lib/ransack/adapters/active_record/context.rb
@@ -212,9 +212,9 @@ module Ransack
             end
           when Arel::Nodes::Equality
             pk = primary_key
-            if join_root.left == pk
+            if join_root.left.eql?(pk)
               join_root.right
-            elsif join_root.right == pk
+            elsif join_root.right.eql?(pk)
               join_root.left
             else
               nil


scarroll32:
> [!NOTE]

Thanks @javieromar7 — excellent diagnosis. The deprecation warning is the visible symptom, but the correctness bug underneath it is the real find: once `ArelExtensions` redefines `==` to return an `Arel::Nodes::Equality` node, `if join_root.left == pk` is testing an object that is never `nil` or `false`, so the branch is unconditionally taken and `join_root.right` comes back whether or not the left side is actually the primary key. That's the kind of bug that produces quietly wrong joins rather than an exception.

`eql?` is the right substitute: `Arel::Attributes::Attribute` is a `Struct`, so `eql?` compares relation and name by value and isn't part of the surface ArelExtensions overrides.

CI is green across all 25 checks. Verified locally merged into current `main` — Rails 8.1.3 / Ruby 3.4.9 / SQLite: **526 examples, 0 failures, 1 pending**.

Issue #1680, opened 2026-09-21
Replace Docusaurus with Just the Docs


scarroll32:
> [!NOTE]
> Issue opened by **Claude** (Claude Code), acting on behalf of @scarroll32.

Implementation issue for the documentation half of #1613.

## Why

The docs site is a Docusaurus 2 app under `docs/`. It pulls in React 17, webpack, Babel, Algolia search and a large transitive npm tree, none of which relate to what the site actually is: 23 Markdown files and a logo. The cost shows up as a steady stream of Dependabot PRs for transitive dev dependencies that never affect a reader of the site — #1650, #1658, #1662 and #1666 are all open right now, and they are a recurring tax on a queue that is already long.

#1613 proposed dropping the site entirely and redirecting to the README. That loses the per-topic navigation and deep links that people cite in issues. [Just the Docs](https://just-the-docs.com/) keeps both while removing the npm tree: it is a Jekyll theme, so GitHub Pages builds it natively with no `package.json`, no lockfile and no build step of our own.

It is also a better fit for this repo — Ransack is a Ruby project, and a Jekyll theme is maintained with `bundle update` alongside everything else rather than a parallel JavaScript toolchain.

## What exists today

| | |
| --- | --- |
| Generator | Docusaurus 2 (`docs/package.json`, `docs/docusaurus.config.js`, `docs/sidebars.js`, `docs/babel.config.js`) |
| Content | 23 Markdown files under `docs/docs/` — 6 in `getting-started/`, 16 in `going-further/`, plus `intro.md` |
| Blog | one post, `docs/blog/2022-03-27-ransack-3.0.0.md` |
| Custom code | `docs/src/` — a React homepage component and two CSS modules |
| Assets | `docs/static/img`, `docs/static/logo` |
| Search | `@easyops-cn/docusaurus-search-local` |
| Build | `.github/workflows/deploy.yml` (Node 16, `yarn build`, `peaceiris/actions-gh-pages`) and `.github/workflows/test-deploy.yml` |

## Proposed work

- [ ] Add a Just the Docs scaffold: `docs/_config.yml`, `docs/Gemfile`, minimal `docs/index.md`
- [ ] Port the 23 Markdown files; the front matter changes from Docusaurus's `sidebar_position` to Just the Docs' `nav_order` / `parent`, and the two sections become parent pages
- [ ] Rewrite the React homepage as a Markdown landing page
[...]

PR #1681, opened 2026-09-21
Publish to RubyGems from GitHub Releases via Trusted Publishing

scarroll32:
> [!NOTE]

Answers "can we push to RubyGems from GitHub using releases?" — yes, and this wires it up.

## What this does

Adds `.github/workflows/release.yml`, which pushes the gem when a GitHub Release is published. It uses [RubyGems Trusted Publishing](https://guides.rubygems.org/trusted-publishing/): the job exchanges a short-lived GitHub OIDC token for a RubyGems credential via `rubygems/release-gem@v1`, so there is **no API key stored in the repo** and nothing to rotate.

The release process becomes: bump `version.rb` → publish a GitHub Release → the gem appears on RubyGems. The manual `gem signin` / `rake build` / `gem push` dance is kept in the docs as a fallback.

## The version guard

Before pushing, the workflow compares the release tag to `Ransack::VERSION` and fails if they disagree:

```
Tag v5.0.0 does not match Ransack::VERSION (4.4.1). Refusing to publish.
```

This is the failure mode worth guarding: publishing a release tagged `v5.0.0` from a tree that still says `4.4.1` yields a gem whose version silently contradicts its tag, and RubyGems releases can't be recalled.

## One-time setup needed before this works

Two things only a gem owner / repo admin can do:

1. **RubyGems.org** — add a trusted publisher at https://rubygems.org/gems/ransack/trusted_publishers:

   | Field | Value |
   | --- | --- |
   | Repository owner | `activerecord-hackery` |
   | Repository name | `ransack` |
   | Workflow filename | `release.yml` |
   | Environment | `rubygems` |

2. **GitHub** — create a `rubygems` environment in [repository settings](https://github.com/activerecord-hackery/ransack/settings/environments). Empty is fine. Adding required reviewers to it means every release needs a second maintainer to approve the push — recommended for a gem with ~100M downloads, but optional.

Until step 1 is done the workflow will run and fail at the push step; nothing else is affected.

## Notes

- `id-token: write` is scoped to the one job that needs it; the workflow default is `contents: read`.
- `workflow_dispatch` is included so a failed publish can be retried against an existing tag without cutting a new release.
- Docs updated at `docs/docs/going-further/release_process.md`.

Relates to #1640 — worth having in place before 5.0.0 ships.


scarroll32:
> [!NOTE]

All three Copilot findings were correct. Rather than patch around them I've replaced the publish mechanism, because findings 1 and 3 share a root cause: `rubygems/release-gem` runs `bundle exec rake release`, and `rake release` is a *source release* task, not a publish task.

I checked [the action's definition](https://github.com/rubygems/release-gem/blob/v1/action.yml) — its final steps are `bundle exec rake release` and `gem exec rubygems-await`. `Bundler::GemHelper` defines `release` as build → `guard_clean` → `source_control_push` → `rubygem_push`, so it wants to create and push the version tag itself. That's the right shape when a tag push *starts* the release; it's the wrong shape here, where a published GitHub Release means the tag already exists and the checkout is detached. It also drags in the full `bundle install` that finding 1 identified — and this Gemfile builds Rails from git, so that's a slow and fragile dependency for a step that only needs to build a gemspec.

**What it does now:** [`rubygems/configure-rubygems-credentials`](https://github.com/rubygems/configure-rubygems-credentials) — the same action `release-gem` uses internally for the OIDC exchange, pinned to the SHA that `release-gem` itself pins — followed by an explicit `gem build ransack.gemspec` and `gem push`. No bundle, no tag creation, no `contents: write`, and the whole thing is three legible steps.

**Finding 2** (shell injection via the `workflow_dispatch` tag input) is fixed as suggested: the value goes through the step `env:` and is expanded as `"$TAG"` rather than being interpolated into the script body. This mattered — the job holds `id-token: write`.

Verified locally: version extraction returns `4.4.1` and `gem build ransack.gemspec` produces `ransack-4.4.1.gem`.

---

Unrelated, but the push surfaced it: GitHub reports **136 Dependabot vulnerabilities** on `main` (2 critical, 64 high). Essentially all of them are in `docs/yarn.lock` — the Docusaurus tree. #1680 removes that tree entirely, which would clear the security tab in one change.

scarroll32:
> [!NOTE]

Second round of Copilot findings — both correct, both fixed.

**Unqualified ref on the dispatch path.** `actions/checkout` resolves a bare name as a branch before it looks for a tag, so dispatching with `v4.5.0` would publish a *branch* of that name in preference to the release tag. That's a genuinely nasty failure mode for a retry path: you'd be republishing from whatever that branch happened to point at. The input is now qualified:

```yaml
ref: >-
  ${{ github.event.inputs.tag
      && format('refs/tags/{0}', github.event.inputs.tag)
      || github.ref }}
```

The `release` path already supplies `github.ref`, which is fully qualified.

**Mutable action tags in an OIDC-enabled job.** `actions/checkout@v4` and `ruby/setup-ruby@v1` are moving targets, and this job can mint a RubyGems publishing token. Both are now pinned to reviewed commits, consistent with how `configure-rubygems-credentials` was already pinned:

| Action | Pin |
| --- | --- |
| `actions/checkout` | `11d5960` (v4.4.0) |
| `ruby/setup-ruby` | `a0102e0` (v1.324.0) |
| `rubygems/configure-rubygems-credentials` | `dc5a8d8` (v2.1.0) |

All three SHAs resolved from the upstream release tags via the API rather than copied from anywhere.

---

### To use this for the 4.5.0 release

The workflow is ready, but Trusted Publishing needs **one manual step that only a gem owner can do**, and it has to happen before the release is published:

1. Go to https://rubygems.org/gems/ransack/trusted_publishers and add a GitHub Actions publisher:

   | Field | Value |
   | --- | --- |
   | Repository owner | `activerecord-hackery` |
   | Repository name | `ransack` |
   | Workflow filename | `release.yml` |
   | Environment | `rubygems` |

2. Create a `rubygems` environment at https://github.com/activerecord-hackery/ransack/settings/environments. Empty is fine; adding required reviewers means a release needs a second maintainer to approve the push.

Then merge this and #1684 (Version 4.5.0), publish the release tagged `v4.5.0`, and the gem goes out on its own.

If step 1 hasn't happened when the release is published, the workflow fails at the final `gem push` and nothing else is affected — the tag and GitHub Release are already made by then, so you can register the publisher and re-run via `Run workflow` with `v4.5.0`.

Issue #1669, opened 2026-04-04
Ruby 4.0 warnings: method redefinitions, unused variables, mismatched indentation


fedegl:
## Description

Running on Ruby 4.0.1 with Rails 8.1.3, ransack 4.4.1 produces several warnings. All functionality works correctly — these are cosmetic but noisy in test output.

### 1. Method redefined: `arel_predicate` (`lib/ransack/nodes/condition.rb`)

```
lib/ransack/nodes/condition.rb:291: warning: method redefined; discarding old arel_predicate
lib/ransack/nodes/condition.rb:220: warning: previous definition of arel_predicate was here
```

A placeholder `arel_predicate` (raises "not implemented") is defined at line 220, then the real implementation replaces it at line 291.

### 2. Method redefined: `conditions` (`lib/ransack/nodes/grouping.rb`)

```
lib/ransack/nodes/grouping.rb:29: warning: method redefined; discarding old conditions
```

`attr_reader :conditions` at line 4 is superseded by `def conditions` at line 29 (which lazy-initializes `@conditions`).

### 3. Duplicate `attr_reader`: `arel_visitor` (`lib/ransack/context.rb`)

```
lib/ransack/context.rb:7: warning: method redefined; discarding old arel_visitor
```

`attr_reader :arel_visitor` appears on both line 5 and line 7.

### 4. Unused variables (`lib/ransack/adapters/active_record/context.rb`)

```
lib/ransack/adapters/active_record/context.rb:301: warning: assigned but unused variable - stashed_association_joins
lib/ransack/adapters/active_record/context.rb:360: warning: assigned but unused variable - reflection
```

### 5. Method redefined: `value` (`lib/ransack/helpers/form_builder.rb`)

```
lib/ransack/helpers/form_builder.rb:10: warning: method redefined; discarding old value
actionview-8.1.3/lib/active_model/naming.rb:36: warning: previous definition of value was here
```

Monkey-patch on `ActionView::Helpers::Tags::Base#value`. The existing `TODO` comment notes this needs a better solution.

### 6. Mismatched indentation (`lib/polyamorous/activerecord/reflection.rb`)

```
lib/polyamorous/activerecord/reflection.rb:6: warning: mismatched indentations at 'else' with 'if' at 4
```

The `if` at line 4 has 1-space indent instead of 2.

## Environment

- Ruby 4.0.1
- Rails 8.1.3
- ransack 4.4.1

scarroll32:
[...]

Issue #172, opened 2012-12-07
@searchresults are prepopulated with data before a search is done?


richlewis14:
Apologies if this is a simple question

I want to search the recipes within my recipe model, which I can but with a few snags

What I want to be able to do is have the search form in my recipe/index action and then display the results AFTER my search has been completed, possibly an AJAX request here.

 My current setup is below like but even though I have my search results in a block, without even doing a search all my recipes are displayed and then when i complete my search the results are then filtered per my search, so if i choose lasagne then only two results show (As there are 2 lasagne recipes)

Recipe controller

```
def index
@q = Recipe.search(params[:q])
@searchresults = @q.result(:distinct => true)
@q.build_condition
end
```

Index action View

```
<%= render 'shared/searchrecipes' %>
 <h1>Results</h1>
  <table>
   <tr>
     <th>Name Of Dish</th>
     <th>Country Of Origin</th>
     <th>Difficulty</th>
     <th>Preperation Time</th>
     <th>ingredient</th>
     <th>Preperation</th>
     <th>Image</th>
   </tr>

   <% @searchresults.each do |r| %>
    <tr>
     <td><%= r.dish_name %></td>
     <td><% if r.country %><%= r.country.name %></td><% end %>
     <td><%= r.difficulty %></td>
     <td><%= r.preperation_time %></td>
     <td><%= ingredient_names(r.ingredients) %></td>
     <td><%= preperation_steps(r.preperations) %></td>
     <td><%= image_tag r.avatar.url(:thumb)%></td>
    </tr>
    <% end %>
   </table>
```

My Search Form

```
<%= search_form_for @q do |f| %>
<%= f.label :dish_name_cont, "Enter Dish Name" %>
<%= f.text_field :dish_name_cont %>

<%= f.label :country_name_cont, "Choose Country" %>
<%= f.collection_select :country_id_eq, Country.all, :id, :name, multiple: true, :prompt => "Please Select" %>

<%= f.label :difficulty_cont, "Choose Difficulty" %>
<%= f.select :difficulty_eq, options_for_select([['Beginner'],['Intermediate'],['Expert']]),{:prompt => 'Please Select'}%>

<%= f.label :preperation_time_cont, "Choose Preperation Time" %>
<%= f.select :preperation_time_eq, options_for_select([['15..30'],['30..60'],['60..120']]), {:prompt => 'Please Select'}%>

 <%= f.submit "Search", :class => "searchbutton" %>
 <% end %
```

PR #1685, opened 2026-09-21
Require sqlite3 >= 2.9.5 for CVE-2026-54619

scarroll32:
> [!NOTE]

Closes the last open Dependabot alert ([#172](https://github.com/activerecord-hackery/ransack/security/dependabot/172), [GHSA-28hh-pr2h-2w89](https://github.com/advisories/GHSA-28hh-pr2h-2w89), CVE-2026-54619).

## The alert

sqlite3 `<= 2.9.4` has a use-after-free when the same SQLite function name is redefined with a different arity or text encoding. Low severity, and here it's a test-only dependency — sqlite3 is in the `Gemfile`, not the gemspec, so it never reaches anyone installing the gem.

It was worth fixing anyway because it was the **only** alert outside the Docusaurus tree, and both branches of the Gemfile permitted a vulnerable version:

```ruby
if ::Gem::Version.new(rails_version) > ::Gem::Version.new('7.3')
  gem 'sqlite3', '>= 2.1'    # permits <= 2.9.4
else
  gem 'sqlite3', '~> 1.4'    # permits <= 2.9.4, and can never be patched
end
```

## Why the split had to go rather than be bumped

The fix landed in **2.9.5**. The 1.x line ended at **1.7.3** and never received a backport, so the `~> 1.4` branch could not be moved to a patched version at all — it had to leave 1.x.

That branch is live in CI: `test.yml` includes `rails: v7.2.2`, and `"7.2.2" > "7.3"` is false, so the Rails 7.2 jobs were the ones pinned to 1.x.

So the question was whether Rails 7.2 tolerates sqlite3 2.x. **It does.** I ran the full suite locally against both:

| Active Record | sqlite3 | Result |
| --- | --- | --- |
| 7.2.2 (the exact version CI pins) | 2.9.6 | 529 examples, 0 failures, 1 pending |
| 7.2.3.2 | 2.9.6 | 529 examples, 0 failures, 1 pending |
| 8.1.3.1 | 2.9.6 | 529 examples, 0 failures, 1 pending |

Since the gemspec already requires `activerecord >= 7.2`, the conditional has no remaining purpose — every supported Rails version works with 2.x. One constraint replaces it.

## Incidental cleanup

Removing the conditional left `rails_version` computed and never read. That's one of the unused-variable warnings catalogued in #1669, so it goes with it. `rails` itself is still used by the `case` that selects the Active Record source.

## Docusaurus alerts

Separately, the other **135 alerts** — all in `docs/yarn.lock` — have been dismissed as `not_used`, each with a comment recording why: they are build-time dependencies of the documentation site, the gemspec packages only `README.md`, `LICENSE` and `lib/**/*` so none of them ship, and the whole tree is being removed in #1680.

That takes the security tab from **136 open alerts (2 critical, 64 high)** to zero once this merges.


scarroll32:
> [!NOTE]

CI caught something I should have anticipated. The four failures were all Ruby **3.1.4** legs:

```
sqlite3 (>= 2.9.5) was resolved to 2.9.6, which depends on Ruby (>= 3.2)
```

**sqlite3 dropped Ruby 3.1 in 2.9.0. The CVE fix is in 2.9.5.** So there is no patched sqlite3 for Ruby 3.1, and there will not be one — the two constraints are mutually exclusive.

Ruby 3.1 now pins to the last line that supports it, and everything else gets the fix:

```ruby
gem 'sqlite3', RUBY_VERSION >= '3.2' ? '>= 2.9.5' : '~> 2.8.1'
```

I want to be straight about what that means rather than present it as a clean fix: **the Ruby 3.1 CI leg is now knowingly running an unpatched sqlite3.** It's a test-only dependency and the advisory is low severity, so the practical risk is negligible — but it's the wrong shape to stay in, and it will recur as more of the ecosystem drops 3.1.

The real fix is dropping Ruby 3.1, which has been EOL since March 2025. Opened #1686 for 5.0.0, which retires this conditional entirely.

PR #1693, opened 2026-09-21
Read grouping attributes back through ransack_alias

### LINKED ISSUE #689: Ransack form search field value not persisted between requests when using alias

scarroll32:
> [!NOTE]
> Rebased from @DougEdey-Filescom's #1529.

Closes #689. Closes #1529.

## The bug

A form field built from an aliased attribute writes its value under the **real** attribute name but reads it back under the **alias**. `read_attribute` looks up only the name it's given, finds nothing, and the field comes back empty on the next request — so a search form loses that input every time it re-renders.

```ruby
ransack_alias :daddy, :parent_name

search.parent_name_cont  # => "John"
search.daddy_cont        # => nil     ← the form reads this one
```

## The fix

`read_attribute` falls back to resolving the alias when the name it was given holds nothing: strip the predicate, ask the context for the aliased attribute, retry under the real name.

@DougEdey-Filescom's note that it's "only used if we don't have an immediately matching attribute" is the important property — the fallback is on the `nil` path, so the common case does no extra work.

> I'm not sure if there's some unforeseen consequences I'm unaware of

None found. The whole suite is green, including `form_helper_spec` and `form_builder_spec`, which are where a change like this would show up.

## Tests

Their spec, tidied for indentation and given a clearer name. Verified both ways — it returns `nil` instead of `"John"` against `main`'s lib, and passes here.

Full suite: **530 examples, 0 failures, 1 pending** (Rails 8.1.3 / Ruby 3.4.9 / SQLite).

> I looked at the contributing guide but there's nothing about updating the changelog in the PR :)

Correct, and still true — since 4.4.0 the changelog is generated from GitHub Release notes, so PRs don't touch `CHANGELOG.md`.

## Related

#916 and #1520 describe the same underlying problem from different angles. #1525 also targets aliases in forms, with a broader change across 11 files; this narrower fix is worth landing on its own first.


scarroll32:
> [!NOTE]

**Correction: the first push of this branch was missing the fix.** CI failed on every job because the branch contained the new spec but not the `read_attribute` change. My error — while proving the spec fails without the fix, I stashed `lib/`, and the stash restored it unstaged, so the commit took only the spec. The PR description's "verified both ways" was true of my local tree, not of what I pushed.

Rebuilt from `main` at `3c28aeb9` with both files. Re-verified, this time through a real form rather than only the `Grouping` node:

```ruby
q = Person.ransack(daddy_cont: "John")
q.daddy_cont                                   # => "John"
search_form_for(q) { |f| f.text_field :daddy_cont }
# <input type="text" value="John" name="q[daddy_cont]" id="q_daddy_cont" />
```

The `value="John"` is the thing #689 was missing. Full suite 534 examples, 0 failures; CI will confirm.

Issue #1421, opened 2023-05-08
Ransack does not distinguish between "*_cont" for LIKE and "*_i_cont" for ILIKE with PostgreSQL


dailson-igo:
Ransack version: 4.0.0

The documentation doesn't agree with what ransack does. Despite the existence of predicates, "*_cont" for LIKE and "*_i_cont" for ILIKE, for PostgreSQL it treats both as one thing, including the result expected by the test
[predicate_spec.rb](https://github.com/activerecord-hackery/ransack/blob/main/spec/ransack/predicate_spec.rb).

I believe it shouldn't be like that, since there may be performance differences between LIKE and ILIKE, in addition to diverging from the documentation.

Looking at the documentation at [using-predicate](https://activerecord-hackery.github.io/ransack/getting-started/using-predicates/) which reports the following result for the "_cont" predicate
```
>> User.ransack(first_name_cont: 'Rya').result.to_sql
=> SELECT "users".* FROM "users" WHERE ("users".."first_name" LIKE '%Rya%')
```

And the matching page at [search-matches](https://activerecord-hackery.github.io/ransack/getting-started/search-matches/) which reports the following match for the predicate "_cont"

```
Predicate: *_cont
Description: Contains value
Notes: use LIKE
```

When I run the query in my environment with PostgreSQL, the result is different:

```
irb> Common::Person.ransack(name_cont: 'Assem').result.to_sql
=> "SELECT \"common_people\".* FROM \"common_people\" WHERE \"common_people\".\"name\" ILIKE '%Assem%'"
```

agrberg:
Just noticed this myself so thank you for making the issue so I know I'm not imaging things 😆 

FWIW I almost always want case insensitive searching but w/ this bug there doesn't appear to be an option for case sensitive search using Ransack should I need it.

chris-hewitt:
Seeing this on Oracle too. The specs capture the expected behavior for [cont](https://github.com/activerecord-hackery/ransack/blob/v4.3.0/spec/ransack/predicate_spec.rb#L172) and [i_cont](https://github.com/activerecord-hackery/ransack/blob/v4.3.0/spec/ransack/predicate_spec.rb#L214), but `UPPER("...") LIKE UPPER('%...%'))` is what we end up with even for `cont`.

PR #1716, opened 2026-09-21
Add i_start and i_end, the case-insensitive start and end predicates

### LINKED ISSUE #1715: 6.0.0 feedback: cont/start/end became case-sensitive on PostgreSQL, and start/end have no i_ variant to move to
Testing `main` against a Rails 8.1 / PostgreSQL app as asked in #1640. Everything passes — no query raises, and our whole suite is green on `main@105b2c9`. This is a report about the one behaviour change that a green suite cannot show, and a request that follows from it.

Thank you for #1421, by the way — the documentation always said `cont` was `LIKE`, and it now is.

## The change, measured

4.4.1 and earlier: `Ransack::Nodes::Condition` passed no case-sensitivity argument to Arel's `matches`, and Arel defaults to case-insensitive, so on PostgreSQL `cont` compiled to `ILIKE`. On `main`, `condition.rb` computes `case_sensitive = !predicate.case_insensitive` and passes it through. The predicate definitions in `constants.rb` are unchanged; only the plumbing is.

```ruby
# 4.4.1                       # main@105b2c9
User.ransack(email_cont: "a") # LIKE  '%a%' ESCAPE '\'   (was ILIKE)
User.ransack(email_start: "a")# LIKE  'a%'   ESCAPE '\'   (was ILIKE)
User.ransack(email_end: "a")  # LIKE  '%a'   ESCAPE '\'   (was ILIKE)
User.ransack(email_i_cont: "a") # ILIKE '%a%' ESCAPE '\'  (unchanged)
```

`not_cont`, `cont_any` and `cont_all` follow `cont`.

## Why it may be worth a line in the 6.0.0 notes

v5.0.0's release notes have a **Breaking changes** section and it is excellent — the `ESCAPE` change is called out clearly. This one is not in it, because #1421 is on `main` for 6.0.0 rather than in 5.0.0.

The reason I think it earns a prominent mention is the blast radius through ActiveAdmin. `ActiveAdmin::Inputs::Filters::StringInput` declares:

```ruby
filter :cont, :eq, :start, :end
```

`cont` is the default, and that input backs every string column filter in an ActiveAdmin app. So on upgrade, every admin text search in every ActiveAdmin-on-PostgreSQL app silently becomes case-sensitive. Nothing raises; operators just stop finding records they used to find, and there is no error to trace it back from.

## The request: `i_start` and `i_end`

`cont` has somewhere to go — `i_cont`. `start` and `end` do not:

```ruby
Ransack::Predicate.names.grep(/\Ai_/)
# => ["i_cont", "i_cont_all", "i_cont_any"]
```

So after this change, **"begins with, case-insensitively" cannot be expressed in Ransack at all.** Before it, `start` did exactly that on PostgreSQL by accident; now nothing does. That feels like a gap the fix opened rather than an intended narrowing, especially since `i_cont` exists and establishes the naming.

Would you consider adding `i_start`, `i_end` and their `not_` / `_any` / `_all` variants for 6.0.0? They would mirror `i_cont` exactly — `case_insensitive: true` plus a `downcase` formatter — so it looks like a `constants.rb` addition rather than a change to the fix.

Happy to open a PR for it if that would help; say the word and I will follow the stack convention in #1640.

## Not asking for a revert

To be explicit: I am not asking for `cont` to go back to `ILIKE`. The new behaviour matches the documentation, the distinction is real, and the performance argument in #1421 stands. Downstream apps that want case-insensitive search should say so with `i_cont` — they just need `i_start` and `i_end` to be able to.

---

Environment: Ruby 3.4.9, Rails 8.1.3.1, PostgreSQL, `ransack` at `main@105b2c96`, ActiveAdmin 4.0.0.beta22.

scarroll32:
> [!NOTE]

Closes #1715. Part of #1640.

## Why

Since #1421 (on `main` for 6.0.0) `cont`, `start` and `end` render as `LIKE` on PostgreSQL, so they are case-sensitive there. `cont` has `i_cont` to move to; `start` and `end` had no case-insensitive form, so "begins with, ignoring case" could no longer be expressed in Ransack at all. ActiveAdmin's string filter offers exactly `cont`, `start` and `end`, which is how #1715 found it.

## What

`i_start`, `i_end`, their `not_` opposites and the `_any` / `_all` compounds. They mirror `i_cont` exactly: `case_insensitive: true` plus a downcasing formatter, which is `ILIKE` on PostgreSQL and `LOWER(col) LIKE` elsewhere. Four entries in `DERIVED_PREDICATES`; the compounds come from `add_predicate` like every other predicate.

```ruby
Ransack::Predicate.names.grep(/i_/).sort
# => ["i_cont", "i_cont_all", "i_cont_any", "i_end", "i_end_all", "i_end_any",
#     "i_start", "i_start_all", "i_start_any", "not_i_cont", "not_i_cont_all",
#     "not_i_cont_any", "not_i_end", "not_i_end_all", "not_i_end_any",
#     "not_i_start", "not_i_start_all", "not_i_start_any"]

Person.ransack(name_i_start: 'eRN').result.to_sql
# PostgreSQL: ... WHERE "people"."name" ILIKE 'ern%' ESCAPE '\'
# SQLite:     ... WHERE LOWER("people"."name") LIKE 'ern%' ESCAPE '\'
```

No change to the #1421 fix itself; `cont`, `start` and `end` stay `LIKE`.

## Docs

- *Using Predicates*: `i_start` and `i_end` sections next to `start` and `end`
- *Search Matchers*: the twelve new rows in the table (and the two `not_end_any` / `not_end_all` rows that had empty descriptions); the case-sensitivity section lists the `i_` family
- *Upgrading*: the 6.0 note now names `start` and `end` alongside `cont`, points to `i_start` / `i_end`, and says that nothing raises when a filter turns case-sensitive, with ActiveAdmin's string filter as the example to check

## Verification

- 686 examples, 0 failures on SQLite (Rails 8.1.3); 688 on PostgreSQL 18; 686 on Rails 7.2.2.1
- The wildcard-escaping shared example now takes the formatter's pattern as a third argument, so the `start` / `end` shapes are covered by it too
- A row-level example creates a person and checks `i_start` / `i_end` find it with different case, and that plain `start` / `end` do not on PostgreSQL
- RuboCop clean; `jekyll build` under `docs/` succeeds


scarroll32:
_Review by Claude (Claude Code), on behalf of @scarroll32._

Reviewed this properly rather than trusting it because I opened it. **The implementation is correct.** One test could not fail for the bug it guards; that is fixed in b4bc8f4 and described at the end.

## What I verified rather than took on trust

**All six forms really do render case-insensitively.** Registering the predicates against `main` and printing the SQL:

```
i_start          : "users"."first_name" ILIKE 'er%' ESCAPE '\'
i_start_any      : ("users"."first_name" ILIKE 'er%' ESCAPE '\' OR  ... ILIKE 'ri%' ...)
i_start_all      : ("users"."first_name" ILIKE 'er%' ESCAPE '\' AND ... ILIKE 'ri%' ...)
not_i_start      : "users"."first_name" NOT ILIKE 'er%' ESCAPE '\'
not_i_start_any  : (... NOT ILIKE 'er%' ... OR  ... NOT ILIKE 'ri%' ...)
not_i_start_all  : (... NOT ILIKE 'er%' ... AND ... NOT ILIKE 'ri%' ...)
```

This was worth checking because of an asymmetry in Arel:

```
matches_any:          [[:req, :others], [:opt, :escape], [:opt, :case_sensitive]]
does_not_match_any:   [[:req, :others], [:opt, :escape]]
```

`does_not_match_any` / `_all` take **no** `case_sensitive` argument, so the `not_` compounds were the plausible place for a silent fallback to `LIKE`. `Condition` already handles it at `condition.rb:445-449` by mapping them over `does_not_match`, so they are fine — but only because of that special case, which is why they now have examples of their own.

**The mechanism is inherited, not reimplemented.** `case_insensitive: true` is the whole contract: `condition.rb:294` wraps the column in `LOWER()` when the dialect has no `ILIKE`, and `:439` derives Arel's flag from the same predicate. Setting the flag genuinely is all that is required.

**Nothing in `lib/` special-cases `i_cont`**, so there is no second registration point these could have missed. Grepped.

**`.downcase` before escaping is safe for non-strings.** The formatter calls `v.downcase` on the value, which looks like a `NoMethodError` waiting for an integer — but Ransack casts to string before the formatter runs. Tested with `1`, a Symbol and an Array against `cont`, `i_cont` and `i_start`; all three behave identically, and the new predicates introduce no new exposure.

**CI is green across the whole matrix** — sqlite, postgres, mysql and trilogy on Rails 7.2, 8.0 and 8.1.

## The finding, now fixed

`it 'works with the _any and _all compounds'` asserted:

```ruby
expect(@s.result.to_sql).to match /I?LIKE 'er%'.* OR .*I?LIKE 'ri%'/
```

`I?` accepts `LIKE` and `ILIKE` equally, so a compound that fell back to case-sensitive `LIKE` on PostgreSQL would still have passed. Checked against a synthetic regressed string rather than reasoned about:

```
correct (ILIKE) matches regex?   true
REGRESSED (LIKE) matches regex?  true
```

It was the only assertion covering compounds, and the compounds are the risky part. **b4bc8f4** splits it per dialect the way the neighbouring examples already do, adds the `not_` compounds, and separates the `Predicate.names` registration check into its own example.

Proof it now bites: removing `case_insensitive: true` from `i_start` alone fails **three** examples. Against the previous version it failed one — and not either compound.

690 examples on SQLite, 692 on PostgreSQL, 0 failures.

## One thing deliberately left alone

`i_start_all` and `not_i_start_any` are close to vacuous in practice — a string cannot begin with two different prefixes, so the first is unsatisfiable and the second is a tautology. But `start_all` and `not_start_any` already have exactly that shape, so generating the full compound set is the consistent choice and a special case here would be the surprise.

PR #1729, opened 2026-09-22
Fix out-of-range integer inside an In node or a composite condition

agrberg:
`Condition#arel_predicate` unwraps an oversized `Casted` integer value into
a plain integer, to avoid `ActiveModel::RangeError` when Arel renders it.
The check only looked at a single `Arel::Nodes::Binary` node's `right`
value, so it missed two shapes the same predicate can build.

An `In` or `NotIn` node holds an Array of `Casted` values in `right`, so
an `_in` predicate with an out-of-range value still raised:

```ruby
Person.ransack(salary_in: ['99999999999999999999']).result.to_sql
# ActiveModel::RangeError: 99999999999999999999 is out of range for
# ActiveRecord::ConnectionAdapters::SQLite3Adapter::SQLite3Integer
# with limit 8 bytes
```

Combining conditions with `And`/`Or`/`Grouping` nests the `Binary` one
level deeper, so an otherwise-fine `_eq` on the same column raised as
soon as it was combined with another condition:

```ruby
Person.ransack(m: 'or', salary_eq: '99999999999999999999',
                name_eq: 'Bob').result.to_sql
# same RangeError, even though salary_eq alone does not raise
```

This replaces the single-node check with a recursive walk over the node
shapes `#arel_predicate` can produce (`Grouping`, `And`, `Or`, `In`,
`NotIn`, `Binary`), so both cases are unwrapped the same way the
original single-`Binary` case already was.

Every built-in predicate's generated SQL is unchanged. Full suite passes
on SQLite, PostgreSQL, MySQL and Trilogy.

No CHANGELOG entry: `CHANGELOG.md` says changes since 4.4.0 are captured
in [Release info](https://github.com/activerecord-hackery/ransack/releases)
instead.


agrberg:
🤦 Sorry folks, Claude missed my instruction to wait for me to review the code AND PR description before opening it. Feel free to give it a look and I look forward to CI's results but I will also explicitly say when I've reviewed the output.

agrberg:
I've reviewed and this makes sense so I will open it up for official review.

The `tl;dr` I'd offer is, in the situation where a user supplies a value too large for an integer column, we'd rather this still go to the SQL query where it is highly likely no values would be returned (except in the `OR` case with other criteria) rather than it raise and abort the whole chain.

